PR #26, opened 2021-07-18
Fix title of column "title" in places list

lenzls:
The column "title" in the places list view always showed an empty column header.
This fixes this to show the correct header "Title".


DavidMStraub:
Thanks, good catch!

PR #70, opened 2022-04-22
Docker compose for easier development

bsudy:
Added Docker Compose setup for easier development.
The frontend and the api are running in separate docker containers. In order to run them on a single port there is reverse proxy added. The reverse proxy forwards all `/api` request to the `gramps-api` container and all other requests go to the `gramps-frontend` container.

The frontend is running in a dev server that makes development without rebuild possible. It also enables auto reload in the browser.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+node_modules
--- a/dev/Dockerfile.api
+++ b/dev/Dockerfile.api
@@ -1,0 +1,16 @@
+FROM dmstraub/gramps-webapi:latest-devel
+
+# copy config file
+COPY config.cfg /app/config/
+
+# extract & use Gramps example tree
+ENV GRAMPS_RESOURCES /usr/share/
+RUN cp -r /usr/share/doc/gramps/example/gramps/example.gramps.gz .
+RUN gunzip example.gramps.gz
+RUN gramps -C Example -i example.gramps --config=database.backend:sqlite --config=database.path:/app/.gramps/grampsdb
+RUN rm -rf appcache
+
+ENV HOME /app
+
+# build full-text search index
+RUN python3 -m gramps_webapi  --config /app/config/config.cfg search index-full
--- a/dev/Dockerfile.api
+++ b/dev/Dockerfile.api
@@ -13,4 +13,12 @@ RUN rm -rf appcache
 ENV HOME /app
 
 # build full-text search index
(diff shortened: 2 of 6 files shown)


DavidMStraub:
Hi @bsudy, thanks a lot for the efforts, but are you aware of Gramps Web, particularly the documentation below? I think this is where all deployment stuff should be handled/documented.

https://gramps-project.github.io/web/Deployment/

bsudy:
Hi,

Thanks for pointing me to the other repo. The purpose of this PR is not to deploy something to production but to make development easier. Instead of running everything locally the whole stack can be run in docker and start developing very fast.

DavidMStraub:
Ok thanks! I'll have a look

DavidMStraub:
Yep, looks great and will make frontend development easier :rocket: 

Ideas:

- Wouldn't we want to make the Gramps DB directory a volume? That way, changes to the DB could persist, which can be useful e.g. for adding some objects that test features of the frontend that are not present in the example DB. The developer can always delete the volume to have a fresh start
- Always developing with `DISABLE_AUTH` seems a bit limited, e.g. we cannot work on the frontend of the user system. We could remove `DISABLE_AUTH` from the cfg and instead use an evironment variable in the compose file. We could add one user per auth level in the backend docker file, something like

```
python3 -m gramps_webapi user add --role 4 owner owner
python3 -m gramps_webapi user add --role 3 editor editor
python3 -m gramps_webapi user add --role 2 contributor contributor
python3 -m gramps_webapi user add --role 1 member member
python3 -m gramps_webapi user add --role 0 guest guest
python3 -m gramps_webapi user add --role -1 disabled disabled
```
(and adding the `users.sqlite` to a volume just like in the normal Gramps Web deployment docs)
etc.

Devs could then still decide switching auth on/off with a simple switch in the compose file.

bsudy:
Great suggestion. I'll take a look and implement it. Thanks

DavidMStraub:
Hi @bsudy, are you still planning to do this? Do you need help? Would be great to get this over the finish line.

DavidMStraub:
@bsudy I just implemented the suggested changes. I think it's ready to merge, let me know in case you have comments.

DavidMStraub:
@bsudy have you found a way to make the "watch" option work, i.e. have the browser reload in case of code changes?

DavidMStraub:
Got it working with 028edfe80b8294a12d26a73df664ed5c14631c6c!

PR #85, opened 2022-08-13
fix code formatting usign npm run format:prettier

khashashin:


DavidMStraub:
Sorry, but I don't want to have a commit changing 166 files just due to formatting.

It's not that everything is unformatted. I am just using a different formatter in VS Code (I think eslint). It's also possible that not all files use the same formatting convention as I changed it at some point.

I think it's better to impose a style by convention and then enforce it e.g. for modified files.

khashashin:
@DavidMStraub sorry, I've just asked in https://web.libera.chat/#gramps before this PR.

do you use this code? https://github.com/gramps-project/Gramps.js/blob/main/package.json#L4

DavidMStraub:
Hi, interesting, I'm not using Libera, maybe I should :)

Will get back to you on the second question. Let's open an issue for it.

PR #87, opened 2022-08-15
Feat/i18n

khashashin:


DavidMStraub:
Looks great! Let me know when it's ready to test.

khashashin:
Well, it's actually ready to test :). If it works this way, I would add other languages. And these json files could already be processed by weblate.

DavidMStraub:
In Firefox 103.0, during development I'm getting "Uncaught SyntaxError: import assertions are not currently supported". So development would require Chrome, but in production (built), it would work cross browser?

khashashin:
> but in production (built), it would work cross browser?

Well I'm not familiar with rollup, it seems you are using rollup to build the application instead of `npm run build`. So, I would suggest to build this version in usual way you did before and check if it works.

DavidMStraub:
Confirming development mode works on Chomium...

DavidMStraub:
> it seems you are using rollup to build the application instead of npm run build

`npm run build` runs rollup:
https://github.com/gramps-project/Gramps.js/blob/main/package.json#L13

DavidMStraub:
So far I am failing to run `npm install`, getting errors like

> npm WARN tarball tarball data for test262@git+ssh://git@github.com/tc39/test262.git#47ab262658cd97ae35c9a537808cac18fa4ab567 (sha512-/vnRSe83+0IwuzYel6XEx/O6moIYc2uylR4YVntxSM/N8Xb0oGKZHs0P3eWGyPee/OyvVObyys14sMHht24cww==) seems to be corrupted. Trying again.

khashashin:
> So far I am failing to run `npm install`, getting errors like

Might be the version of Node, I'm using 16+. But let me do some more checks on this PR because I can't build the project using `npm run build`. Getting error:

```
╰─ npm run build

> gramps-js@0.13.3 build
> rimraf dist && rollup -c rollup.config.js


./index.html → dist...
Browserslist: caniuse-lite is outdated. Please run:
  npx browserslist@latest --update-db
  Why you should do it regularly: https://github.com/browserslist/browserslist#browsers-data-updating
(node:2536) [DEP0148] DeprecationWarning: Use of deprecated folder mapping "./" in the "exports" field module resolution of the package at /mnt/c/Users/khash/Development/gramps-project/check_front/Gramps.js/node_modules/tslib/package.json.
Update this package.json to use a subpath pattern like "./*".
(Use `node --trace-deprecation ...` to show where the warning was created)
[!] (plugin import-assertions) TypeError: self.load is not a function
src/lang/index.js
TypeError: self.load is not a function
    at /mnt/c/Users/khash/Development/gramps-project/check_front/Gramps.js/node_modules/rollup-plugin-import-assertions/src/index.js:143:16
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Promise.all (index 1)
    at Object.transform (/mnt/c/Users/khash/Development/gramps-project/check_front/Gramps.js/node_modules/rollup-plugin-import-assertions/src/index.js:126:7)
    at ModuleLoader.addModuleSource (/mnt/c/Users/khash/Development/gramps-project/check_front/Gramps.js/node_modules/rollup/dist/shared/rollup.js:19708:30)
    at ModuleLoader.fetchModule (/mnt/c/Users/khash/Development/gramps-project/check_front/Gramps.js/node_modules/rollup/dist/shared/rollup.js:19761:9)
    at async Promise.all (index 0)
    at ModuleLoader.fetchStaticDependencies (/mnt/c/Users/khash/Development/gramps-project/check_front/Gramps.js/node_modules/rollup/dist/shared/rollup.js:19792:34)
    at async Promise.all (index 0)
```

However, I do not get any error message when installing the packages. I tried to install the packages by cloning the repository to a new folder (reinstall)

khashashin:
It's strange... When I run directly `rimraf dist && rollup -c rollup.config.js` (you need install the cli's globally using `-g`) it builds the project correctly without any errors but when I run `npm run build` it fails.

khashashin:
Now `npm run build` works without errors, see: https://github.com/swiing/rollup-plugin-import-assertions/issues/3

khashashin:
> but in production (built), it would work cross browser?

Just tested the production version in firefox and the translation works without errors:

https://user-images.githubusercontent.com/17837758/185169827-449a8718-54fa-44c5-8fdb-3551c15e89fa.mp4

DavidMStraub:
Perfect! Ready to merge?

khashashin:
> Perfect! Ready to merge?
Yes, just added few translations. What should be done with translation files for other languages? Do they have to be created as well?

DavidMStraub:
The problem with the first run assistant is that it runs before the user chooses user settings, so the translations will have no effect. It also won't work because before creating an admin account, we cannot fetch the strings from the API. So this would require major changes.

DavidMStraub:
> What should be done with translation files for other languages? Do they have to be created as well?

That depends on how the integration with Weblate works, which I don't know. I assume they will be populated by commits from Weblate, but no idea if they have to be created first. But I suggest to tackle Weblate integration in a separate PR.

khashashin:
> The problem with the first run assistant is that it runs before the user chooses user settings, so the translations will have no effect. It also won't work because before creating an admin account, we cannot fetch the strings from the API. So this would require major changes.

Should I do revers commit? Doesn't `this._()` method show english strings by default?

DavidMStraub:
Yes, it will show the English strings, but I'm not sure it makes sense to have the strings translated if it cannot be used.

What we could consider is setting the langauge in the first run assistant without any user interaction automatically to the browser's default language. After completing the assistant and when showing the user settings dialog, we could preselect that language in the language selection dialogue.

The backend strings would not be available at the time of the first run assistant, but that shouldn't matter as there are probably only "frontend" strings in this form.

But perhaps it makes more sense to do that in a separate PR.

khashashin:
> What we could consider is setting the langauge in the first run assistant without any user interaction automatically to the browser's default language. After completing the assistant and when showing the user settings dialog, we could preselect that language in the language selection dialogue.

Yeah sounds as a solution and we could define fallback language as english.

> But perhaps it makes more sense to do that in a separate PR.

So, should we leave the strings in this PR?

DavidMStraub:
> So, should we leave the strings in this PR?

If you plan to implement translation of the first run assistant, I think it's fine to leave them in.

khashashin:
> > So, should we leave the strings in this PR?
> 
> If you plan to implement translation of the first run assistant, I think it's fine to leave them in.

Yes, I'll do it in separate branch

DavidMStraub:
Can you please add to the frontend development docs in the web repository that it now works only in Chrome during development?

DavidMStraub:
Can you please add to the frontend development docs in the web repository that it now works only in Chrome during development?

DavidMStraub:
Can you please add to the frontend development docs in the web repository that it now works only in Chrome?

khashashin:
> Can you please add to the frontend development docs in the web repository that it now works only in Chrome?

Yes

PR #88, opened 2022-08-19
Load frontend strings in connectedCallback

khashashin:
Detect browser language and set it as Application lang.

image

Firefox settings are in Russian. My Windows System Language is English

image


DavidMStraub:
The class constructor should not contain any logic but only be used for static variable initialization.

https://lit.dev/docs/components/lifecycle/

A better place would be in connectedCallback.

DavidMStraub:
Some additional comments:

- What is `navigator.userLangauge`?
- When the browser language doesn't match a translation, it looks like you are setting the language to `en` but are not showing the language selection dialog on first login. Why is that? Shouldn't we offer the user the choice, as we do now, if the autodetection fails?
- We need to handle composed language names. For instance, for Portuguese, there is no `pt` in Gramps but only `pt_BR` and `pt_PT`. So I would suggest to first try matching the full language code (replacing `-` with `_`), then only the part before `-` (as you do now). Perhaps we would even want to match the first part against the beginning of a Gramps language code, such that a user with, say, `zh-sg` would end up with one of the other Chinese options present in Gramps.

khashashin:
> * What is `navigator.userLangauge`?

It was a code copied from [Stackoverflow](https://stackoverflow.com/a/44114093/7986808) with slight adjustments. `navigator.userLanguage` seems to be an [api for IE](https://stackoverflow.com/a/1043451/7986808). But since it is no longer maintained, I think we can just use `navigator.language` to get the browser language

> * When the browser language doesn't match a translation, it looks like you are setting the language to en but are not showing the language selection dialog on first login. Why is that?

I've miss interpret our dialog from here: https://github.com/gramps-project/Gramps.js/pull/87#issuecomment-1219880280 + https://github.com/gramps-project/Gramps.js/pull/87#issuecomment-1220078539.  So I've to leave this part, right?

```
if (!getSettings().lang) {
  this.loadingState = LOADING_STATE_MISSING_SETTINGS
}
```

> * We need to handle composed language names. For instance, for Portuguese, there is no pt in Gramps but only pt_BR and pt_PT.

I will delegate this logic to a separate function. I will update the PR.

khashashin:
> * We need to handle composed language names. For instance, for Portuguese, there is no pt in Gramps but only pt_BR and pt_PT. So I would suggest to first try matching the full language code (replacing - with _), then only the part before - (as you do now). Perhaps we would even want to match the first part against the beginning of a Gramps language code, such that a user with, say, zh-sg would end up with one of the other Chinese options present in Gramps.

I think we could just let the user choose the language they prefer, as it has been until now. The last commit handles the matching of the existing languages and if defined in the string returns it, otherwise it returns null, so we could ignore the browser language identification since we can't display it anyway.

DavidMStraub:
I agree we don't need the complicated logic, but in your new version you're only matching the full string, right? So it will not work for most of the languages in Gramps which are two-character strings. Or am I missing something?

khashashin:
> I agree we don't need the complicated logic, but in your new version you're only matching the full string, right? So it will not work for most of the languages in Gramps which are two-character strings. Or am I missing something?

You're right, so we need to check the full match first and if it fails then try to match the first part, right? For instance if the user has one of the following browser language set en-IE, en-IN, en-NZ, en-US, en-ZA, en-GB we should try to match it full which will fail because we don't have these keys in our frontend (but we have the en_GB.po file in the backend, which confuses me a bit because it doesn't match the frontend key), and then we will look for the first part "en" that matches our frontend strings.

---

This matching process only goes through the frontend strings, right? I don't need to check which strings I received from the backend? It is not 100% clear to me for what purpose are used following list in `strings.js`:

```
export const grampsStrings = [
  '_Media Type:',
  '_Names',
  '_Save',
  'Abbreviation',
  'Abort changes?',
  'About',
  ...,
]
```

DavidMStraub:
> so we need to check the full match first and if it fails then try to match the first part, right?

Yes, exactly.

> but we have the en_GB.po file in the backend, which confuses me a bit because it doesn't match the frontend key

Yes, the `en_GB` locale should have a corresponding key in the frontend strings, while the frontend `en` key corresponds to a US locale. That this is not named `en_US` probably has historical reasons and the source of truth here is [this dictionary](https://github.com/gramps-project/gramps/blob/master/gramps/gen/utils/grampslocale.py#L70-L116) in Gramps.

> This matching process only goes through the frontend strings, right? I don't need to check which strings I received from the backend?

Right, we match with the keys in the frontend strings simply because we do the matching before we have fetched the strings from the backend; for instance, during the first run assistant, we don't have a token that allows us to fetch backend strings.

>  It is not 100% clear to me for what purpose are used following list in strings.js:

This is just a list of all the strings that are needed - in principle the Web API allows to translate any string translated in Gramps, but we are fetching only those that are needed in Gramps.js. We do this once and cache them rather than making an API call for every single string to be translated.

src/GrampsJs.js
```diff
@@ -582,6 +578,13 @@ export class GrampsJs extends LitElement {
     window.addEventListener('db:changed', () => this._loadDbInfo(false))
     this.addEventListener('drawer:toggle', this._toggleDrawer)
     window.addEventListener('keydown', (event) => this._handleKey(event))
+
+    const browserLang = this._getBrowserLanguage()
+    if (browserLang) {
+      updateSettings({ lang: browserLang })
```
DavidMStraub, line 588: What you are doing here, I think, is always overriding the user's settings with the browser's default language. This is not nice as the user might prefer to use a different language from the browser (e.g. I sometimes work on computers with English locale but prefer my native language for the app). The point of persisting the settings in localStorage is for the users to set this permanently.

So I think the logic should be that browserLang is only used to update the settings if the language is not already set.

src/GrampsJs.js
```diff
@@ -582,6 +578,13 @@ export class GrampsJs extends LitElement {
...
+      this.settings = getSettings()
+      this._loadStrings(grampsStrings, this.settings.lang)
```
DavidMStraub: You cannot do this in connectedCallback as the user might not be logged in or even registered, so the call will fail.

The original call you removed was done as part of the render method at a point where it's guaranteed we are already authenticated.

src/GrampsJs.js
```diff
@@ -430,9 +429,6 @@ export class GrampsJs extends LitElement {
     if (this.loadingState === LOADING_STATE_MISSING_SETTINGS) {
       return this._renderOnboarding()
     }
-    if (this.settings.lang && Object.keys(this._strings).length === 0) {
-      this._loadStrings(grampsStrings, this.settings.lang)
-    }
```
DavidMStraub: See comment below; that we are doing the call here makes sense as we know we are authorized at this point. However I do see the issue that `Object.keys(this._strings).length === 0` won't work as we have already loaded the frontend strings...

src/GrampsJs.js
```diff
@@ -729,6 +743,14 @@ export class GrampsJs extends LitElement {
...
 
   _loadStrings (strings, lang) {
+    // First load the strings for the current language
+    if (lang in additionalStrings) {
+      this._strings = additionalStrings[lang]
+      this._strings.__lang__ = lang
+      this._lang = lang
+    }
+
+    // Then load the strings from the server and merge them with the current language strings
```
DavidMStraub: If I see it correctly, you are adding the additionalStrings twice, no?

So far this method was doing the API call for the backend strings and, after it succeeded, added the additonalStrings. Now we want a logic that allows already using the additionalStrings before we can make an API call. I see two options to implement it: either using the `if ('error' in data)` part of the fetch handler. Currently, it shows an error message (`_showError`) but we could also say if there is an error (e.g. since we are not authorized yet), just load the additionalStrings. The second possibility would be to do something like the `this.loadingState > LOADING_STATE_UNAUTHORIZED_NOCONNECTION` that is done in the `update` method: only fetch the backend strings if we know we are authorized, otherwise just load the additionalStrings.

src/GrampsJs.js
DavidMStraub: On second thought (see discussion below), if we rewrite `loadStrings` to fall back to frontend strings only in case of no authorization we can do it, but we have to make sure backend strings are loaded once we are authorized.

src/GrampsJs.js
khashashin: Yes you're right its bad ux, I didn't think about it in this way. I'll change it to:

```javascript
  connectedCallback () {
    super.connectedCallback()
    this._loadDbInfo()
    window.addEventListener('db:changed', () => this._loadDbInfo(false))
    this.addEventListener('drawer:toggle', this._toggleDrawer)
    window.addEventListener('keydown', (event) => this._handleKey(event))

    const browserLang = this._getBrowserLanguage()
    if (browserLang && !getSettings().lang) {
      updateSettings({ lang: browserLang })
      this.settings = getSettings()
      this._strings = additionalStrings[browserLang]
      this._strings.__lang__ = browserLang
      this._lang = browserLang
    }
  }
```

khashashin:
Just to summarize your comments: https://github.com/gramps-project/Gramps.js/pull/88#discussion_r952775929, https://github.com/gramps-project/Gramps.js/pull/88#discussion_r952809534, https://github.com/gramps-project/Gramps.js/pull/88#discussion_r952777764 and https://github.com/gramps-project/Gramps.js/pull/88#discussion_r952788530, we don't have to try to get the strings from the backend when we are not logged in.

1. Since the front-end strings are here (they are not required to be downloaded from anywhere) we can handle all this logic independently from `this._loadStrings()` method. This means for me that it can be written in `connectedCallback()` lifecycle.
2. We only need to get the strings from the backend when we are logged in, so it makes no sense to call the `this._loadStrings()` method at all without checking if we are logged in.

I think the first point is easy to solve. I will simply do it in the following way:

```javascript
  connectedCallback () {
    super.connectedCallback()
    this._loadDbInfo()
    window.addEventListener('db:changed', () => this._loadDbInfo(false))
    this.addEventListener('drawer:toggle', this._toggleDrawer)
    window.addEventListener('keydown', (event) => this._handleKey(event))

    const browserLang = this._getBrowserLanguage()
    if (browserLang && !getSettings().lang) {
      updateSettings({ lang: browserLang })
      this.settings = getSettings()
      this._strings = additionalStrings[browserLang]
      this._strings.__lang__ = browserLang
      this._lang = browserLang
    }
  }

  _getBrowserLanguage () {
    // get browser language and replace all '-' with '_'
    // since the strings from backend comes with underscore
    const browserLang = navigator.language.replaceAll('-', '_')
    console.log('Browser language: ', browserLang)
    if (browserLang in additionalStrings) {
      return browserLang
    } else if (browserLang.split('_')[0] in additionalStrings) {
      return browserLang.split('_')[0]
    } else {
      return null
    }
  }
```

The second point is a bit tricky (I just haven't understood the whole flow of views yet). What if we were not able to define the browser language? What would happen if we call `this._loadStrings(grampsStrings, this.settings.lang)` and `this.settings.lang` is undefined? Or do I not understand the flow correctly and it is not even possible that `this.settings.lang` can be undefined?

Can I call `this._loadStrings(grampsStrings, this.settings.lang)` in the `renderContent()` method without worrying as follows:

```javascript
  renderContent () {
    if (this.loadingState === LOADING_STATE_INITIAL) {
      return this._renderInitial()
    }
    if (this.loadingState === LOADING_STATE_UNAUTHORIZED_NOCONNECTION) {
      return this._renderNoConn()
    }
    if (this.loadingState === LOADING_STATE_UNAUTHORIZED) {
      window.history.pushState({}, '', 'login')
      return this._renderLogin()
    }
    if (this.loadingState > LOADING_STATE_UNAUTHORIZED_NOCONNECTION) {   // <-------------
      this._loadStrings(grampsStrings, this.settings.lang)   // <-------------------------------
    }   // <--------------------------------------------------
    if (this.loadingState === LOADING_STATE_NO_OWNER) {
      window.history.pushState({}, '', 'firstrun')
      return this._renderFirstRun()
    }
    if (!getSettings().lang) {
      this.loadingState = LOADING_STATE_MISSING_SETTINGS
    }
    if (!getSettings().homePerson && this._dbInfo?.object_counts?.people) {
      // show settings dialog for missing home person only if the db is not empty!
      this.loadingState = LOADING_STATE_MISSING_SETTINGS
    }
    if (this.loadingState === LOADING_STATE_MISSING_SETTINGS) {
      return this._renderOnboarding()
    }
```

DavidMStraub:
Hi,
sorry that this file is a big mess - it should be made more modular at some point. I think there are still a couple of things missing. But I have an idea, can we simplify this PR to just define a new function `getBrowserLanguage` (doesn't have to be a class method and could even be in `utils.js` and a new method, something like `_loadFrontendStrings(lang)` containing basically the snipped you have in connectedCallback? Then in a second step I can come up with a PR that hopefully handels the flow correctly.

DavidMStraub:
And in `connectedCallback` you can already put the call as you have it now,

```python
const browserLang = getBrowserLanguage()
if (browserLang && !this.settings.lang) {
    updateSettings({ lang: browserLang })
    this._loadFrontendStrings()
}
```

DavidMStraub:
No, sorry, actually it needs to be 

```python
const browserLang = getBrowserLanguage()
if (browserLang && !this.settings.lang) {
    updateSettings({ lang: browserLang })
}
if (this.settings.lang) {
    this._loadFrontendStrings()
}
```

That way, somebody who is logged out but had `lang` set in localStorage, will see a translated login form for instance.

khashashin:
I think this PR is now much cleaner and trying to solve individual problems :). Should be now much easier to review

DavidMStraub (approved):

PR #124, opened 2022-10-25
Added custom event type to form select with form validation

Airick73:
After discussing this implementation with @DavidMStraub in this issue: https://github.com/gramps-project/Gramps.js/issues/118 . I am submitting this pull request to add a custom event type feature.

### DIFF
--- a/src/components/GrampsjsFormSelectType.js
+++ b/src/components/GrampsjsFormSelectType.js
@@ -27,6 +27,7 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
       required: {type: Boolean},
       initialValue: {type: String},
       noheading: {type: Boolean},
+      _hasCustomType: {type: Boolean} //adding _hasCustomType prop
     }
   }
 
@@ -43,6 +44,7 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
     this.required = false
     this.initialValue = ''
     this.noheading = false
+    this._hasCustomType = false //this will be false as no custom type is entered initially
   }
 
   getTypes(types) {
@@ -65,7 +67,8 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
       <p>
         <mwc-select
           style="width:100%"
-          ?required="${this.required}"
+          ?required=${!this._hasCustomType}
+          ?disabled=${this._hasCustomType}
           validationMessage="${this._('This field is mandatory')}"
           @change="${this.handleChange}"
           ?disabled="${this.disabled}"
@@ -91,7 +94,45 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
               )}
         </mwc-select>
       </p>
+
+      <mwc-button
+        @click="${this.switchTypeInput}"
+      >
+        Enter custom type instead
+      </mwc-button>
+
+      ${this._hasCustomType === false
+        ? html``
+        : html`
+          <h4 class="label">${this.heading || this._('Custom Type')}</h4>
+          <p>
+          <mwc-textfield
+            style="width:100%"
+            ?required=${this._hasCustomType}
+            validationMessage="${this._('This field is mandatory')}"
+            @change="${this.handleChange}"
+            id="custom-type"
+          >
+          ${types.map(
+                  (obj, i) => html`
+                    <mwc-list-item
+                      value="${typesLocale[i]}"
+                      ?selected="${(this.initialValue &&
+                        obj === this.initialValue) ||
+                      obj === this.defaultTypeName}"
+                      >${this._(obj)}</mwc-list-item
+                    >
+                  `
+                )}
+            </mwc-textfield>
+          </p>
+        `
+      }
     `
+  }
+
+  switchTypeInput(){
+    this._hasCustomType = true
   }
 
   reset() {
@@ -115,7 +156,8 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
 
   isValid() {
     const selectType = this.shadowRoot.getElementById('select-type')
-    if (selectType === null) {
+    const customType = this.shadowRoot.getElementById('custom-type') //adding query for custom-type id
+    if (selectType === null && customType === null) { //checking if both types null then return false
       return false
     }
     try {
--- a/src/components/GrampsjsFormSelectType.js
+++ b/src/components/GrampsjsFormSelectType.js
@@ -2,12 +2,12 @@
 Element for selecting a Gramps type
 */
 
-import {html, LitElement} from 'lit'
+import { html, LitElement } from 'lit'
 import '@material/mwc-select'
 import '@material/mwc-list/mwc-list-item'
 
-import {sharedStyles} from '../SharedStyles.js'
-import {GrampsjsTranslateMixin} from '../mixins/GrampsjsTranslateMixin.js'
+import { sharedStyles } from '../SharedStyles.js'
+import { GrampsjsTranslateMixin } from '../mixins/GrampsjsTranslateMixin.js'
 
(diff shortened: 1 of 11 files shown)


src/components/GrampsjsFormSelectType.js
```diff
@@ -85,9 +87,47 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
               )}
         </mwc-select>
       </p>
+
+      <mwc-button
+        @click="${this.switchTypeInput}"
+      >
+        Enter custom type instead
```
DavidMStraub: Please wrap this in `this._('Enter ...')` and add the string to `en.json` so it can be translated.

Would "add custom type instead" perhaps be more explicit that this will permanently create a new type, what do you think?

src/components/GrampsjsFormSelectType.js
```diff
@@ -85,9 +87,47 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
...
+      </mwc-button>
+
+      ${this._hasCustomType === false
```
DavidMStraub: Can be `!this._hasCustomType`

src/components/GrampsjsFormSelectType.js
```diff
@@ -109,7 +149,8 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
 
   isValid() {
     const selectType = this.shadowRoot.getElementById('select-type')
-    if (selectType === null) {
+    const customType = this.shadowRoot.getElementById('custom-type') //adding query for custom-type id
+    if (selectType === null && customType === null) { //checking if both types null then return false
```
DavidMStraub: They will only be `null` if the elements don't exist. Is that your intention?

src/components/GrampsjsFormSelectType.js
```diff
@@ -85,9 +87,47 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
...
+      <mwc-button
+        @click="${this.switchTypeInput}"
+      >
+        Enter custom type instead
+      </mwc-button>
+
+      ${this._hasCustomType === false
+        ? html``
+        : html`
+          <h4 class="label">${this.heading || this._('Custom Type')}</h4>
```
DavidMStraub: `this.heading` is a way to override the default `Type` label of the type selector. It doesn't make sense to override "Custom Type` with the same heading, does it?

src/components/GrampsjsFormSelectType.js
```diff
@@ -85,9 +87,47 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
...
+            style="width:100%"
+            ?required=${this._hasCustomType}
+            validationMessage="${this._('This field is mandatory')}"
+            @change="${this.handleChange}"
+            id="custom-type"
+          >
+          ${types.map(
+                  (obj, i) => html`
+                    <mwc-list-item
+                      value="${typesLocale[i]}"
```
DavidMStraub: Why do you have a list of `mwc-list-item`s inside the textfield?

src/components/GrampsjsFormSelectType.js
Airick73: I like 'Enter custom type instead' or just 'Enter custom type'. Imo 'add' refers to something being additional which is not the case, we are switching the input to custom type. I would also consider 'Switch to custom type' as we are changing the input option.

I've made these code changes and will implement with your feedback on the naming.

image

src/components/GrampsjsFormSelectType.js
Airick73: Changing `${this._hasCustomType === false` to `${!this._hasCustomType` . I agree this is cleaner.

src/components/GrampsjsFormSelectType.js
Airick73: My intention was to check if both types were empty:

**if custom type AND select type do not contain input
&nbsp;&nbsp;  return the form is not valid**

Before it seemed that the `isValid()`  function was only checking if select type was empty:

**if select type does not contain input 
&nbsp;&nbsp;  return the form is not valid**

src/components/GrampsjsFormSelectType.js
Airick73: This was just a misunderstanding of the architecture. I thought that `${this.heading || this._('Custom Type')}` was required in the event a heading was not able to generate. I agree it is not necessary. I've changed it to 
`${this._('Custom Type')}`. 

_Note: I am also adding "Custom Type" to en.json_

src/components/GrampsjsFormSelectType.js
Airick73: It is redundant and I've replaced the `mwc-list-list` from inside the textfield with `${this.types}`. I was modeling the way that select type was saving the input to `types`. I still have a hard time understand the code inside `mwc-select` . It seems that having `${this.types}` inside my textfield will function as intended to save the custom input to types. 
image

Airick73:
Thank you for your feedback in this pull request. It has been very informative and continues to help me in better understand the Gramps.js architecture. I've made the necessary changes to the code and I would like to re-submit a pull request for your approval.

src/components/GrampsjsFormSelectType.js
DavidMStraub: Thanks, makes sense!

src/components/GrampsjsFormSelectType.js
DavidMStraub: Right, but `selectType === null` does not check whether the form contains valid input. The existing code
```python
if (selectType === null) {
      return false
    }
```
is only there because when the element has not fully rendered yet, `#select-type` will not exist in the shadow root yet and checking `selectType.validity.valid` will raise an error.

src/components/GrampsjsFormSelectType.js
DavidMStraub: No, I don't think there should be anything inside the `mwc-textfield` element. Do you see an effect when you remove it?

The way the type is saved is that `handleChange` is called when you edit the textfield. This fires a `formdata:changed` event which is handled in `GrampsjsObjectForm.js`.

Airick73: After removing the `${this.types}` the ability to save custom events remains intact. I removed it and left the `mwc-textfield` blank.

src/components/GrampsjsFormSelectType.js
Airick73: So `isValid()` is for checking if the content has rendered properly. It is called in `checkFormValidity()` from `GrampViewNewEvent.js` in this case. 

There are other elements to check in the form such as date, private, description, but only the `select-type` has a required variable attached to it. Is that the reason we are only checking for the rendering of `select-type`? 

If that is the case then the `custom-type` would also be required if the `select-type` was not being used. Or is the intention here just to check that `select-type` was rendered properly and assume if it was rendered properly the form is valid. Wouldn't we also want to make sure the `custom-type` was rendered properly to consider it a valid form?

To conclude if the intention is just to check if `select-type` was rendered then I would agree that the code to check `custom-type` validity should be removed.

DavidMStraub:
I tried this out but it throws an error, I think it is because the `mwc-select` has the `disabled` attribute defined twice with different value.

DavidMStraub:
Thanks, it works now!

Just very minor things left now:

- The "cancel" button calls the `reset` method; it needs to clear the custom type field so we can select a non-custom type again. This can be achieved by just setting `this._hasCustomType = false` in there.
- The "switch to custom type" button remains enabled after clicking it even though it has no effect anymore. I would either disable it when `this._hasCustomType` is true or, perhaps even better, change it to have the opposite effect, namely switching back to a non-custom type.
- I caused a merge conflict by merging in updated translations, sorry about that.

DavidMStraub:
I will make some final changes and force-push in a couple of minutes

PR #148, opened 2022-12-28
Allow editing note type (fixes #122)

### LINKED ISSUE #122: Allow editing note type

DavidMStraub:


src/components/GrampsjsFormEditNoteType.js
```diff
@@ -0,0 +1,35 @@
...
+import {GrampsjsObjectForm} from './GrampsjsObjectForm.js'
+
+class GrampsjsFormEditNoteType extends GrampsjsObjectForm {
+  renderForm() {
+    return html`
+      <grampsjs-form-select-type
+        id="note-type"
+        .strings="${this.strings}"
+        ?loadingTypes="${this.loadingTypes}"
+        ?disabled="${this.loadingTypes}"
```
fuelen: I've removed `disabled` property in https://github.com/gramps-project/Gramps.js/pull/141, as it was not used internally in `grampsjs-form-select-type`. In all cases it had the same value as `loadingTypes` so in updated version the component uses it directly in order to disable select box

DavidMStraub: Ok, let's merge this after #141.

PR #178, opened 2023-02-09
Localization of the date in the pedigree card.

basdl:
The dates shown on the pedigree cards were not localized. This is now done using the toLocaleDateString function of dates.

### DIFF
--- a/src/components/GrampsjsPedigreeCard.js
+++ b/src/components/GrampsjsPedigreeCard.js
@@ -115,11 +115,17 @@ class GrampsjsPedigreeCard extends LitElement {
               <br />
               <span class="dates">
                 ${this.person?.profile?.birth?.date
-                  ? html`* ${this.person.profile.birth.date}`
+                  ? html`*
+                    ${new Date(
+                      Date.parse(this.person.profile.birth.date)
+                    ).toLocaleDateString()}`
                   : ''}
                 <br />
                 ${this.person?.profile?.death?.date
-                  ? html`† ${this.person.profile.death.date}`
+                  ? html`†
+                    ${new Date(
+                      Date.parse(this.person.profile.death.date)
+                    ).toLocaleDateString()}`
                   : ''}
               </span>
               <div class="more">


DavidMStraub:
Thanks! Can you please post screenshots how it looked like for you before and what it looks like with the fix? For me, it is just an ISO date. Trying to understand how much this is a frontend issue vs. backend.

DavidMStraub:
(Since the backend is supposed to localize the strings using Gramps functionality.)

DavidMStraub:
(and a user might choose to run the frontend in a different language than their browser is using.)

basdl:
Okay, got it. So the question is then, how to render different date formats for the current user. I assumed that the dates coming from the backend are locale agnostic, and the browser / client is localizing appropriately. 


Before: 
Screenshot 2023-02-09 123508


After: Screenshot 2023-02-09 122735

DavidMStraub:
There are two "layers" of data in the API: the raw Gramps objects (where "date" is a locale agnostic data structure) and the "profile" view that returns localized, human readable strings, which is used here.

Actually, I realized your approach will not work since dates in Gramps can be more complicated than in the example, you can have different calendars, before/after, approximate, etc. ... fortunately, we don't have to handle all that in Javascript as Gramps already does all that. If you set a date to "estimated before 1755", you will notice that this will be correctly translated when you change the frontend language setting.

So the reason you are seeing ISO dates here is that this is the default setting in Gramps for the date format, even the localized one. There is an alternative setting "numerical" (in Gramps Desktop, see settings > display) which would render as dd.mm.yyyy with German locale.

The actual question is whether "numerical" should be used here in the frontend - I don't think so, but acknowledge that users might disagree. So we could simply add that as a frontend setting.

This feature would require adding the date format as a query parameter in the API; it's relatively easy to implement. But I would close this PR for now.

I hope you don't feel bad about doing the PR in vain. The interplay between Gramps.js and Gramps Web API can be pretty confusing and it's often useful to first open an issue to clarify what needs to be done.

PR #184, opened 2023-02-26
Better estimation of ancestor tree sizes

digitaltembo:
Previously, the dimensions of the Ancestor Tree d3 SVG were estimated as
```
const width = depth * boxWidth + (depth - 1) * gapX + 2 * padding
const height = boxHeight * 2 ** (depth - 1) + (2 ** (depth - 1) - 1) * gapY
```

Which is to say, effectively (depth + gapX) * boxWidth by (boxHeight + gapY) * 2^(depth - 1). This is a reasonable estimate if the ancestor tree is fully populated, but is quite innacurate if the data is more sparse.

This new estimate ensures:
a) that the width is no wider than the total available depth of ancestors available 
b) that the height is representative of the actual height of the nodes rendered 
c) that the tree itself is centered in the SVG, instead of the home person

It would be better to actually inspect what d3 is rendering and get the height/width from there, but I couldn't see a great way of doing that.

Before:

https://user-images.githubusercontent.com/4743325/221389244-b6f0526b-42bd-4baf-94a7-f6aa6661539e.mov

After:

https://user-images.githubusercontent.com/4743325/221389248-8dad4005-e5df-4fb2-b4bb-6d8282622e8e.mov

### DIFF
--- a/src/charts/TreeChart.js
+++ b/src/charts/TreeChart.js
@@ -2,6 +2,31 @@ import {create} from 'd3-selection'
 import {hierarchy, tree} from 'd3-hierarchy'
 import {curveBumpX, link, symbolTriangle, symbol} from 'd3-shape'
 
+// Returns the total depth of the tree
+function countDepthOfTree(treeData) {
+  if (treeData == null) {
+    return 0
+  }
+  return (
+    1 +
+    Math.max(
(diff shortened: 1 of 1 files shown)


DavidMStraub:
Brilliant, thanks!

PR #186, opened 2023-02-26
Better estimation of fan chart size

digitaltembo:
Pretty similar to #184, although here it is a bit easier to get the actual sizes

Instead of assuming that the fan chart is a perfect circle of radius 60 * <ancestor generations> + 30, this calculates the bounding box of the ancestor arcs that are actually drawn.

Before:

https://user-images.githubusercontent.com/4743325/221392267-66532cd2-f0af-4c4c-b75f-797245253c2f.mov

After:

https://user-images.githubusercontent.com/4743325/221392274-c85fdac9-b4b8-431c-9127-b5608bde5a62.mov

### DIFF
--- a/src/charts/FanChart.js
+++ b/src/charts/FanChart.js
@@ -11,30 +11,82 @@ function defaultColor(d) {
   return schemePaired[ind]
 }
 
+// Finds the bounding rectangle of a list of rectangles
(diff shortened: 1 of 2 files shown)


DavidMStraub:
Awesome :rocket:

PR #187, opened 2023-02-26
Add information to map popups

digitaltembo:
This adds the location parent list, an image, any notes and a scrollable list of events that happened in that place to the place-popups on the map:

Screenshot 2023-02-26 at 3 49 22 PM

Haven't really used lit before, so some of the interactions between it and the leaflet API (expecting `popup.setContent(<some HTML string>)`) are a bit awkward - didn't render any lit elements inside the popup, only regular web components, and couldn't figure out how to get css styles to apply to them, so I just used inline-css. Particularly here, I couldn't figure out how to reuse the `<grampsjs-events>` or `<grampsjs-view-object-notes>` components, so this effectively reimplements them a bit within the GrampsjsMapMarker

Basically, this is a tiny bit hacky. While it works quite nicely with my initial setup, I don't know whether it makes sense to put into the main branch

### DIFF
--- a/src/components/GrampsjsMapMarker.js
+++ b/src/components/GrampsjsMapMarker.js
@@ -1,18 +1,28 @@
+import '@material/mwc-circular-progress'
+import '@material/mwc-list'
+
 import {html, LitElement} from 'lit'
 import {mdiMapMarker} from '@mdi/js'
+
 import {Marker, Icon} from '../../node_modules/leaflet/dist/leaflet-src.esm.js'
 import {iconDataUrl} from '../icons.js'
+import {apiGet} from '../api.js'
+import {GrampsjsTranslateMixin} from '../mixins/GrampsjsTranslateMixin.js'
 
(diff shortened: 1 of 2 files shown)


DavidMStraub:
Hi, thanks for the pull request! Actually, I am a bit suprprised you feel the need to add a place popup - in fact there were popups before, but I replaced them with a "more info" box very similar to Google Maps for several reasons; it's easier to make it mobile-friendly and it's easier to add lots of (scrollable) content. So adding back popups on top is redundant in my opinion.

When you click on a pin in Google Maps, it also doesn't open a popup, but the box:

grafik


Gramps Web:

grafik

Adding a list of events (to this box) is of course great, there are just two things to consider:

- We should try to reuse components for a list of events that are already used in the app (probably `grampjs-events`)
- I am not sure how to handle large number of events. For instance, my home town has many hundreds of events and it doesn't make sense to load all of them. Perhaps show the X most recent ones? Needs some thought.

PR #317, opened 2023-11-24
add zoom and pan to charts

geostag:
d3-zoom ist used to add zoom and pan to tree- and fanchart.
To take advantage of intensive zooming, the area to display the chart is expanded to width 100% of the current superseding element and height of slightly less then the window hight. This leads to a somewhat larger display of the chart on usual screen sizes. Please let me know, if this fits into the desired look+feel.


DavidMStraub:
Thanks for the PR, that's indeed a huge improvement! :rocket: 

The only thing I don't like too much is indeed the fact that the font size is often much larger than the rest of the text:

grafik

Perhaps we can find a way to set the default zoom to have the same or smaller font size as the rest of the app.

Another thing that occurred to me looking at other websites with slippy UI (e.g. the trees at FamilySearch, Ancestry, MyHeritage) is that they often have gray background, which seems to signal that it is a scrollable/zoomable canvas. But we can implement that in the future (along with possibly +/-/reset zoom buttons) if we find a way to make it pretty.

DavidMStraub:
Perhaps a simple solution would be to add a scale factor to the chart functions which multiplies height & width in the viewbox, and calculate/estimate the optimal initial value by getting `window.innerHeight` & ``window.innerWidth` in the `connectedCallback` of the custom elements?

geostag:
tried to implement this. `windows.innerHeight` and `windows.innerWidth` are directly accessible from the chart generating js, so getting them in `connectedCallback` seems not to be necessary. After tweaking around a bit, I get better results with just scaling according to width of window and chart instead of the areas, especially with very small charts.

DavidMStraub:
Ok cool! The scaling is still sometimes strange, but since the user can easily zoom in/out now, I think we can leave it like this for now.

Thanks!

PR #335, opened 2023-12-23
Add bookmark feature

DavidMStraub:
- [x] Add bookmark button to each object
- [x] Implement personal bookmark feature for all users using local storage
- [x] Add bookmarks view
- [ ] Implement global/public bookmark storage in Gramps database


DavidMStraub:
Setting as ready for review as we could already release this feature with local-only bookmarks and add the public bookmarks later.

PR #401, opened 2024-03-09
Relationship graph fork

geostag:
this should do the relationship graph, styled in gramps-web manner, as a first version


DavidMStraub:
Finally tried it out, awesome work!

Two things that caught my attention (using the Gramps example database):

- I have three unconnected people. Is that expected or should I investigate?
- We should change the string for max number of ancestor generations to maximum degree of separation
- It's currently not yet possible to click on a person to center it, like for the other charts, but that should be straightfoward to add in analogy, shouldn't it?

image

geostag:
- unconnected people: which person did you choose as root person for the chart to get the disconnected ones?
- agreed. Whats is the best way to do this: overwrite the renderControls in ViewTreeChartbase or create another _setDegSep and trigger on this in the generic renderControls?
- Right, I simply forgot this :-) - will add it.

geostag:
- unconnected pople: I also found situations, where I get unconnected people. The chart builds up nodes from two family sources: extended.primary_parent_family and extended.families. In cases, a person is part of additional extended.families, but neither the partner nor any children are part of the person selection, this creates an orphaned node. I suggest, that I simply check, if any other person in the family is part of the person selection. 
- string "max number of ancestor generation": still not sure, what is the best way to do that. Maybe we also want an "all persons" button, to select all persons, wether they are connected to the center person or not. And thus need a specific "renderControls"?
- clickable person: fixed

DavidMStraub:
> agreed. Whats is the best way to do this: overwrite the renderControls in ViewTreeChartbase or create another _setDegSep and trigger on this in the generic renderControls?

Yes, I think `_setDegSep` (or perhaps more concisely `_setSep`) is the best approach.

geostag:
From my point of view ready for merge. 

Please note: The person nodes do not contain images (different than in the other charts). This chart is intended to contain a very large number of persons, thus pulling images for all of them seems to be a bad idea.

DavidMStraub:
Great work!

I feel the navigation tabs are getting a bit too long and who knows what else we will add :wink:
Should we switch to something more efficient? Text only? Icons with tooltips? Dropdown?

image

andre1wall:
I am experiencing some problems with the correct sorting of the main person's descendants.
maybe i'm the only one experiencing these errors as my family tree is quite large, but it would be good if the descendants were always sorted by year of birth and no distinction was made as to whether the people were married or not, as this should not affect the view.

wrong sorting descendants:
gramps-web-realationship-praph
sorting


a person who has been married twice is reported twice:
the same person was married twice1
the same person was married twice2

thank you for your hard work i really appreciate it.
this view was really missing.
please continue with your great work

Issue #67, opened 2022-02-21
Feature - save family tree view settings


OJ7:
Whenever I go to the family tree view, I usually change it to vertical view with 5 ancestors/descendants. I have to do this each time I open up the website. It would be nice to have this persist across sessions.

DavidMStraub:
Yes, good idea.

If somebody wants to implement this:

`GrampsjsViewGraph` has a property `settings` that is an object that contains all the settings (synced with LocalStorage). So, just need to implement two things:

- set view parameters based on a new settings key, if present
- update this settings key when a parameter is changed, using `updateSettings({someKey: someValue})` from `api.js`

BlitzundBogen:
I could try to solve this. 😄

DavidMStraub:
April 2024 update on the implementation: use `this.appState.settings` and `this.appState.updateSettings` (which are available in all components).

kinslayer1982:
I guess adding the option to change the default values is related to this request. Or should it be a separate one?
For now I changed the default values in the .js file and mount it inside the container as a quick workaround.

DavidMStraub:
The default values are the default values, then there are user settings which could be made persistent.

kinslayer1982:
Well, that's a VERY simple point of view, don't you think? I bet you wouldn't like all software having everything hardcoded without the ability to set reasonable default values. According to your answer nothing like environment variables would exist, no way of customizing the deployment of software, containers, vms...

Let's take the invoicing software I set up for my wife: The default currency of the software is $. This doesn't mean that she has to use $ everywhere. No, I can configure the system to use € as default currency for every invoice that gets created. Which again does not prevent her to set the currency for specific customers to $, €, or whatever...

Yes, you need defaults so the software may run even if nothing is specified at the time of deployment. But it doesn't mean that these defaults are suitable for every situation / use-case.

I'm just offering ideas to you and others that are not comfortable with the hardcoded defaults and no way to change that.
If you don't want that I can stop making suggestions or try and investigate issues I and share what I find.

DavidMStraub:
This issue is about saving preferences for the family tree view. It's completely unnecessary to start an argument about the necessity of preferences. If the user sets their view to use € instead of $, that's a preference, and the default (which is used when the user hasn't set a preference) is one of the two.

kinslayer1982:
[...]

PR #404, opened 2024-03-15
Improve names edit handling (fixes #123) (fixes #149)

### LINKED ISSUE #123: Manage last names
I've imported my database to gramps web and I noticed that I'm not able to manage multiple last names.
It would be great to have this option, as in my case it is one of the most used features.

### LINKED ISSUE #149: Edit Person -> Manage Names: Cannot remove a surname
In my imported database many of the women will have 2 names under "married name", usually one given, one taken. I have been editing this to more correctly use the Birth Name/Married name.

Within the web editor there is no ability to remove a surname from the list of surnames available.

image

I want to delete this surname, but it's not possible.
image

BlitzundBogen:
Added some fixes and improvements in names' edit handling.

Fixes:
- name-type not selected
- fix show-more feature

Features:
- add sort to surname
- add delete to name and surname

UX/UI improvements:
- use editable list for names
- add name edit form validation
- improve name-type vs. custom-type handling
- some minor styling

Fixes:
- https://github.com/gramps-project/gramps-web/issues/149
- https://github.com/gramps-project/gramps-web/issues/123

### DIFF
--- a/src/components/GrampsjsFormSelectType.js
+++ b/src/components/GrampsjsFormSelectType.js
@@ -143,6 +143,9 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
   }
 
   handleChange(e) {
+    if (e.target.disabled) {
+      return
+    }
     const data = e.target.value
     this.dispatchEvent(
       new CustomEvent('formdata:changed', {
--- a/src/components/GrampsjsFormSelectType.js
+++ b/src/components/GrampsjsFormSelectType.js
@@ -2,16 +2,24 @@
 Element for selecting a Gramps type
 */
 
-import {html, LitElement} from 'lit'
+import {css, html, LitElement} from 'lit'
 import '@material/mwc-select'
 import '@material/mwc-list/mwc-list-item'
 
+import {classMap} from 'lit/directives/class-map.js'
 import {sharedStyles} from '../SharedStyles.js'
 import {GrampsjsTranslateMixin} from '../mixins/GrampsjsTranslateMixin.js'
 
 class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
   static get styles() {
-    return [sharedStyles]
+    return [
+      sharedStyles,
+      css`
+        .hide {
+          display: none;
+        }
+      `,
+    ]
   }
 
   static get properties() {
@@ -69,7 +77,8 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
         : html`<h4 class="label">${this.heading || this._('Type')}</h4>`}
       <p style="display: flex">
         <mwc-select
-          style="width:100%; ${this._hasCustomType ? 'display: none' : ''}"
+          style="width:100%"
+          class="${classMap({hide: this._hasCustomType})}"
           ?required=${this.required && !this._hasCustomType && !this.nocustom}
           ?disabled=${this.loadingTypes || this._hasCustomType || this.disabled}
           validationMessage="${this._('This field is mandatory')}"
--- a/src/components/GrampsjsFormSelectType.js
+++ b/src/components/GrampsjsFormSelectType.js
@@ -124,7 +124,6 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
           ? ''
           : html`
               <mwc-icon-button
-                class="edit"
                 style="margin-left: 8px"
                 icon="${this._hasCustomType ? 'remove' : 'add'}"
                 id="button-switch-type"
--- a/src/components/GrampsjsFormSelectType.js
+++ b/src/components/GrampsjsFormSelectType.js
@@ -134,7 +134,7 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
                 for="button-switch-type"
                 content="${this._hasCustomType
                   ? this._('Switch to default type')
-                  : this._('Add custom type')}"
+                  : this._('Switch to custom type')}"
                 .strings="${this.strings}"
               ></grampsjs-tooltip>
             `}
--- a/src/components/GrampsjsFormSelectType.js
+++ b/src/components/GrampsjsFormSelectType.js
@@ -129,6 +129,10 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
   }
 
   switchTypeInput() {
+    if (!this._hasCustomType) {
+      const selectType = this.shadowRoot.getElementById('select-type')
+      selectType.value = null
+    }
     this._hasCustomType = !this._hasCustomType
   }
 
@@ -143,9 +147,6 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
   }
 
   handleChange(e) {
-    if (e.target.disabled) {
-      return
-    }
     const data = e.target.value
     this.dispatchEvent(
       new CustomEvent('formdata:changed', {
@@ -158,13 +159,14 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
 
   isValid() {
     const selectType = this.shadowRoot.getElementById('select-type')
-    const customType = this.shadowRoot.getElementById('custom-type') // adding query for custom-type id
-    if (selectType === null && customType === null) {
-      // checking if both types null then return false
+    const customType = this.shadowRoot.getElementById('custom-type')
+    if (selectType === null || (this._hasCustomType && customType === null)) {
       return false
     }
     try {
-      return selectType?.validity?.valid
+      return this._hasCustomType
+        ? customType?.validity?.valid
+        : selectType?.validity?.valid
     } catch {
       return false
     }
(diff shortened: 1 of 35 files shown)


DavidMStraub:
Great, thanks! Let me know when it's ready for review.

One comment: please don't modify the langugage JSONs, they are updated via Weblate. Unfortunately changing the english string is only possible by adding a new one (unless I'm mistaken).

BlitzundBogen:
> Great, thanks! Let me know when it's ready for review.
> 
> One comment: please don't modify the langugage JSONs, they are updated via Weblate. Unfortunately changing the english string is only possible by adding a new one (unless I'm mistaken).

I was also unsure about the language jsons. I will change it.

Generally I am not a JavaScript crack at all, so I am really happy for feedback for improvements. 😄 
Also I have still issues getting the test locally to run. I guess after fixing this, I might be review ready. I let you know when I am ready.

DavidMStraub:
>  Also I have still issues getting the test locally to run. I guess after fixing this, I might be review ready. I let you know when I am ready.


Unfortunately there are no unit tests for this repo as I'm not good with Javascript unit tests :slightly_frowning_face: . Are you referring to the Web API (Python) tests?

src/components/GrampsjsPerson.js
```diff
@@ -49,16 +49,17 @@ export class GrampsjsPerson extends GrampsjsObject {
...
     const call = this.data?.primary_name?.call
     const callIndex = call ? given.search(call) : -1
     given =
       callIndex > -1
-        ? html`${given.substr(0, callIndex)}<span class="given-name"
-              >${given.substr(callIndex, call.length)}</span
-            >${given.substr(callIndex + call.length)}`
+        ? html`${given.substring(0, callIndex)}<span class="given-name"
+              >${given.substring(callIndex, call.length)}</span
+            >${given.substring(callIndex + call.length)}`
```
DavidMStraub: Good catch to use `substring` in place of `substr`, but on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#the_difference_between_substring_and_substr) I found this

> The two parameters of substr() are start and length, while for substring(), they are start and end.

So don't you have to change the arguments as well?

src/components/GrampsjsPerson.js
```diff
@@ -49,16 +49,17 @@ export class GrampsjsPerson extends GrampsjsObject {
...
         : given
-    return html`${given} ${surname} ${suffix}`
+    return html`${given} ${nick ? `"${nick}" ` : ''}${surname} ${suffix}`
```
DavidMStraub: Would be nice to use typographical quotation marks (although strictly speaking that would even by locale dependent)

DavidMStraub: But on second thought I am not sure I would like to always have the nickname displayed as part of the page title ... does Gramps Desktop do it like that? I think we can use that as guidance to have a more consistent user experience. For instance, if both Gramps & Gramps Web would display nicknames as part of the person profile's title, I would likely want to move more informal nick names to a secondary name to prevent them from showing up...

src/components/GrampsjsFormSelectType.js
```diff
@@ -95,40 +104,49 @@ class GrampsjsFormSelectType extends GrampsjsTranslateMixin(LitElement) {
...
+                icon="${this._hasCustomType ? 'remove' : 'add'}"
+                id="button-switch-type"
+                @click="${this.switchTypeInput}"
+                ?disabled="${this.disabled}"
+              ></mwc-icon-button>
+              <grampsjs-tooltip
+                for="button-switch-type"
+                content="${this._hasCustomType
+                  ? this._('Switch to default type')
+                  : this._('Add custom type')}"
```
DavidMStraub, line 137: I suggest to leave the string as it was orginally to not lose the translations.

DavidMStraub:
I couldn't resist to already add a couple of comments. The code looks really good overall!

src/components/GrampsjsFormSelectType.js
BlitzundBogen: I can do that. Still I would like to propose to add the string 'Add custom type' for this scenario. It's more self explaning. But I woulf leave it like it was for now. 😄

src/components/GrampsjsPerson.js
BlitzundBogen: How do you deal with those kind of suggestions in the project? Should I open a discussion here in gitlab?

But you are right. I guess the best thing would be, if the user could choose (and define it in the user settings). 

The scenario I came up with this was: It used to be like that, that parents and children often share the same given name. Sometimes they differ in second or third given names (which might not be known). To make it easier to distinguish persons, this might help.

I will remove this for now.

In my family tree there are some persons with this issue 😄 This is why they mostly have a nickname. To distinguish e.g. between mother and daughter.

BlitzundBogen:
> > Also I have still issues getting the test locally to run. I guess after fixing this, I might be review ready. I let you know when I am ready.
> 
> Unfortunately there are no unit tests for this repo as I'm not good with Javascript unit tests 🙁 . Are you referring to the Web API (Python) tests?

I haven't really gotten into this by now. I just thought there were some, because there is this "test"-script in the package.json. But okay, then I will ignore this. Maybe I can add some someday 😄

BlitzundBogen:
> Great, thanks! Let me know when it's ready for review.
> 
> One comment: please don't modify the langugage JSONs, they are updated via Weblate. Unfortunately changing the english string is only possible by adding a new one (unless I'm mistaken).

I am still not quite sure how to introduce new strings. I found the `strings.js`, but those translations seem to come from the backend. I saw in one commit a string added to the `en.json`? So this seems to be the way to go!?!

Or does Weblate identifies them inside the code? (I doubt it. 😄 )

src/components/GrampsjsFormSelectType.js
BlitzundBogen: Or maybe I add the string. From UX-perspective I see it like this:
If I set a custom type and save it will become a "default type" (as it is now in the default types list in the frontend). When users read "switch" they might expect to see the custom type set before there.

src/components/GrampsjsPerson.js
BlitzundBogen: Maybe I can implement this when solving https://github.com/gramps-project/gramps-web/issues/67

src/components/GrampsjsFormSelectType.js
DavidMStraub: Not sure what you mean. There are two kinds of types in Gramps: default types and custom types. The default types are translated, while the custom types are just arbitrary strings. You can either select from a list of default types, or use a custom type instead. If you want to replace a custom type by a default type again, you have to switch back to the dropdown list.

DavidMStraub:
Yes, just add new strings to `en.json`.

src/components/GrampsjsFormSelectType.js
DavidMStraub: After trying it out I get your point and I think you're right :smile: I missed the fact that the dropdown contains both custom and default types.

So: great idea.

DavidMStraub:
I tried out your code, great work! It brings the whole name editing experience to a new level. Just two small things:

- The only feature that's still missing is changing the primary name. That would be relatively simple to implement I guess using e.g. a "star" button. Up to you if you want to add it, we can always add it later
- I was confused by the presence of the delete and up/down buttons if the name details are not expanded yet, because they have no effect/meaning:

image


Can you hide them?

BlitzundBogen:
> I tried out your code, great work! It brings the whole name editing experience to a new level. Just two small things:
> 
>     * The only feature that's still missing is changing the primary name. That would be relatively simple to implement I guess using e.g. a "star" button. Up to you if you want to add it, we can always add it later
> 
>     * I was confused by the presence of the delete and up/down buttons if the name details are not expanded yet, because they have no effect/meaning:
> 
> 
> image
> 
> Can you hide them?
 Thanks a lot! And yes!

1. I know remember, I wanted to add chhanging the primary name. Just forgot it last week. 😄 
2. You are right. I will hide the moving options until the details are shown. When you have more names in the list, it's not so bad. But I definitely see the point.

DavidMStraub:
Hi @BlitzundBogen should we merge it as is right now or two you want to implement the two points above? However there is a conflict right now.

BlitzundBogen:
> Hi @BlitzundBogen should we merge it as is right now or two you want to implement the two points above? However there is a conflict right now.

I am sorry, I didn't have the time to improve. I will try to at least solve the conflicts this week. If I do not find the time I will definitely find it the week after. Then I could also do the improvements.

BlitzundBogen:
I had some issues with my dev environment again, which took me some time to solve. For now only the order of names/changing primary name is missing. I would suggest, if you are ready to merge, just do it and I will open a new PR for this issue. If I am faster, i will add it here. But currently I doubt that. 😄

PS: David, what happened to you profile? I cannot find you anymore. 😱

m-breitbach:
I am a big fan of these changes, thank you for taking care. Looking forward to do thorough testing and giving feedback. (I have no dev setup yet, so I will have to wait until the release. Is there documentation for a proper dev setup?)

DavidMStraub:
>  Is there documentation for a proper dev setup

Not sure about the "proper" :wink: but here is the documentation: https://www.grampsweb.org/dev-frontend/setup/

DavidMStraub:
> PS: David, what happened to you profile? I cannot find you anymore. 😱

Yes, Github (or its AI) suspended my account (in error) and it took them one month to process my appeal :sob:

DavidMStraub:
I'm rebasing now.

DavidMStraub:
:tada:

Issue #347, opened 2024-01-10
Integration with Topola Genealogy Viewer


Wdavery:
In desktop Gramps, the [Interactive Family Tree ](https://gramps-project.org/wiki/index.php/Addon:Interactive_Family_Tree) addon ([source](https://github.com/gramps-project/addons-source/tree/master/Topola)) adds an integration with the interactive [Topola Genealogy Viewer](https://github.com/PeWu/topola-viewer).

Including it as an option within the Family Tree page would be valuable, as it provides a great way to view large and complex trees. 
I imagine it would be relatively straightforward to add since the integration with Gramps already exists.

DavidMStraub:
Indeed! I think @PeWu and I were in contact a long time ago - I think this was still the predecessor project of Gramps Web - but it was never implemented. @PeWu would you be willing to support development of such an integration? Contributing would be awesome, but even if you don't have time for it, just being available for asking stupid questions would be helpful. 😄 

In general, our architecture is that we have custom elements, so we would have to wrap the Topola code into a custom element. Using typescript is fine because we are building with rollup.

The main open question for me would be in which format to fetch data. The API can create Gedcom or Gramps XML files, but it seems like an unnecessary complication to generate that in Python and parse it in JS again. The existing charts directly consume the Gramps objects in JSON form.

PeWu:
Hi, can you give me a pointer where to look for the "Family Tree page" you are mentioning? I can definitely help with the integration.

The current way Topola Viewer is integrated with Gramps is that Gramps generates a GEDCOM file that is then parsed in Topola Viewer. This was the easiest way to integrate. Using an API that loads only needed data would be faster but requires some implementation on the Topola side.

My suggestion would be to start with integrating first using Gedcom just to see the result as quickly as possible. Then, we can consider using a more granular API to fetch only needed data.

DavidMStraub:
Sounds good! The component holding the views for the graphs is here:
https://github.com/gramps-project/Gramps.js/blob/main/src/views/GrampsjsViewTree.js

Basically, adding a new view means adding a tab and a new `renderX` method that renders a new component (for isntance `<grampsjs-view-topola>`).

DavidMStraub:
@geostag actually, Topola is quite similar to option 2 in #313 - I don't know if there can be some synergy. In principle, I wouldn't mind if we have several graphs with similar scope, it gives users more choice which is good...

emyoulation:
[...]

PR #414, opened 2024-03-29
Topola Viewer integration (#347)

PeWu:
Topola Viewer is integrated as follows:
- Topola Viewer is loaded from the official deployment at https://pewu.github.io/topola-viewer and displayed in an iframe.
- Data is loaded from the Gramps Web API as a full GEDCOM file and is passed to Topola Viewer using JS messaging between Gramps Web and the iframe. Topola Viewer has a dedicated "embedded mode" for this purpose.

Here are some ideas how to move forward:

1. [recommended] Implement fetching only relevant data through the Gramps Web API instead of fetching the whole GEDCOM. This would have to be mostly implemented on the Topola Viewer side with some plumbing in `GrampsjsViewTopola`.

2. [possible] Instead of showing an iframe, the user could be redirected to a new tab showing Topola Viewer with data from Gramps Web. On one hand, we avoid the combination of 2 apps that have a different look and feel but on the other hand, the integration is far from being seamless.

3. [possible but more effort required] To avoid iframes, we'd have to use the https://github.com/PeWu/topola library directly in Gramps Web. The topola library is responsible for rendering the family tree and doing animations while the Topola Viewer is essentially a rich wrapper around the topola library providing the menu, loading data, showing individual details and settings. Using the topola library directly would mean reimplementing some parts of Topola Viewer in Gramps Web.

### DIFF
--- a/src/GrampsJs.js
+++ b/src/GrampsJs.js
@@ -68,6 +68,7 @@ import './views/GrampsjsViewRecent.js'
 import './views/GrampsjsViewBookmarks.js'
 import './views/GrampsjsViewMap.js'
 import './views/GrampsjsViewTree.js'
+import './views/GrampsjsViewTopola.js'
 import './views/GrampsjsViewNewPerson.js'
 import './views/GrampsjsViewNewFamily.js'
 import './views/GrampsjsViewNewEvent.js'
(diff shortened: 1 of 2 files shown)


DavidMStraub:
Cool, thanks a lot! In principle 3 would be the best option, but we could opt for a staged approach as well. I'll review this version in the next days. :+1:

DavidMStraub:
I'm really sorry that this PR sat here for 2 years now with no action. But I see that topola is now on NPM as a library, so I think integrating it properly should be totally feasible! I'll close this PR in the meantime.

PR #419, opened 2024-04-07
Show page in citations list

stdtom:
As a user I would like to distinguish different citations from the same source in the citations list by the page.


DavidMStraub:
Makes total sense. Nice and easy. Thanks!

stdtom:
My pleasure. Thanks for the quick review and merge.
And by the way: Kudos to you @DavidMStraub  for this awesome project in general... and for the setup of the dev environment in this project in particular. Worked like a charm!

PR #426, opened 2024-04-17
Render GrampsjsFaces even if face_detection fails (fixes #424)

### LINKED ISSUE #424: Not able to edit images if OpenCV is not installed
If OpenCV python bindings are not installed on the server, media objects won't show up in edit mode. In this case, the user can't tag faces. OpenCV is used for face detection, but this feature should not be required.

Zegeri:
These changes avoid a TypeError when face_detection fails and allow any classes extending GrampsjsConnectedComponent to render even if it's in a failed state.


DavidMStraub:
Nice, thanks!

Issue #450, opened 2024-06-10
Picture for each node in RelationshipsTree


JulKoroleva:
Each version of the family tree has an image for each person, but in the **relationship tree** there are no images for the persons when displaying genealogical connections, only names/date/etc. Could you please add the corresponding functionality so that people's photos are displayed there too? Thank you so much!

geostag:
The relationship graph was intended to possibly display a very large part of the gramps project. Thus displaying a possible image for each person might lead to heavy load on server side. However, we could limit the number of images displayed to prevent this situation and thus display the images up to this limit, which should give the desired result in any case, when the number persons is low enough.

geostag:
@DavidMStraub thanks for rebasing and fixing the text length issue. This issue solved with #451

PR #451, opened 2024-06-15
add (limited) images to relship chart

geostag:
#450 limit is set to 50


DavidMStraub:
Let me rebase & force-push due to the changes to translation file handling in main...

DavidMStraub:
Works great!

The only problem I noticed is that long names overflow

grafik

but we can use ellipses like for the other charts

grafik

DavidMStraub:
Should be fixed now

PR #454, opened 2024-06-19
Add OpenHistoricalMap layer to maps

### LINKED ISSUE #437: Add OpenHistoricalMap layer
https://gramps.discourse.group/t/offer-openhistoricalmap-as-an-option-for-the-map-view/5313/12

1ec5:
I’ve added a rudimentary [OpenHistoricalMap](https://wiki.openstreetmap.org/wiki/OpenHistoricalMap) layer to all the maps in Gramps Web. The maps still default to the OpenStreetMap layer, but the layer selection control is now available just about anywhere there’s a map.

For now, the OpenHistoricalMap layer just shows everything that’s in OHM, regardless of the date. This would obviously confuse the user. It would be straightforward to filter the map by date, either by installing the interactive [Leaflet time slider](https://github.com/OpenHistoricalMap/leaflet-ohm-timeslider-v2/) plugin or by passing in a date from the surrounding context. I’m unsure of where to get a relevant date, but it seems like it’s within reach.

Here’s my hometown, Loveland, Ohio, and [the late owner](https://www.wikidata.org/wiki/Q114016441) of a former hardware store there:

The Map view showing Loveland, Ohio, on the OpenHistoricalMap layer.
The Place detail view for Loveland, Ohio, showing the OpenHistoricalMap layer.
The Timeline view for Bessie Sparks, showing her 1895 marriage in Loveland, Ohio, on the OpenHistoricalMap layer.

Fixes #437, working towards [Mantis #13271](https://gramps-project.org/bugs/view.php?id=13271).

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
   "license": "AGPL",
   "dependencies": {
     "@hpcc-js/wasm": "^2.16.0",
+    "@maplibre/maplibre-gl-leaflet": "0.0.21",
     "@material/mwc-button": "^0.27.0",
     "@material/mwc-checkbox": "^0.27.0",
     "@material/mwc-circular-progress": "^0.27.0",
@@ -105,7 +106,7 @@
     "@material/mwc-top-app-bar": "^0.27.0",
     "@material/web": "^1.4.0",
     "@mdi/js": "^7.4.47",
-    "@types/leaflet": "^1.5.19",
+    "@types/leaflet": "^1.9.12",
     "d3": "^7.8.2",
     "d3-array": "^3.2.4",
     "d3-axis": "^3.0.0",
@@ -120,6 +121,7 @@
     "jwt-decode": "^3.1.2",
     "leaflet": "^1.9.4",
     "lit": "^2.1.1",
+    "maplibre-gl": "^4.4.1",
     "pwa-helper-components": "^0.2.10",
     "pwa-helpers": "^0.9.1",
     "tippy.js": "^6.3.7"
--- a/src/components/GrampsjsFormEditLatLong.js
+++ b/src/components/GrampsjsFormEditLatLong.js
@@ -114,6 +114,7 @@ class GrampsjsFormEditLatLong extends GrampsjsObjectForm {
         <grampsjs-map
           latitude="${this.data.lat ? parseFloat(this.data.lat) : 0}"
           longitude="${this.data.long ? parseFloat(this.data.long) : 0}"
+          layerSwitcher
           mapid="edit-latlong-map"
           id="map"
           @click="${this._handleMapClick}"
--- a/src/components/GrampsjsMap.js
+++ b/src/components/GrampsjsMap.js
@@ -58,6 +58,7 @@ class GrampsjsMap extends LitElement {
       _map: {type: Object},
       _glMap: {type: Object},
       _layercontrol: {type: Object},
+      _currentLayer: {type: String},
     }
   }
 
@@ -75,6 +76,7 @@ class GrampsjsMap extends LitElement {
     this.longMin = 0
     this.longMax = 0
     this.layerSwitcher = false
+    this._currentLayer = 'OpenStreetMap'
   }
 
   firstUpdated() {
@@ -99,9 +101,9 @@ class GrampsjsMap extends LitElement {
       zoomControl: false,
     })
     tileLayer.addTo(this._map)
-    const gl = L.maplibreGL({
+    this._gl = L.maplibreGL({
       style: config.glStyle,
-    }).addTo(this._map)
+    })
     this._map.addControl(L.control.zoom({position: 'bottomright'}))
     if (this.locateControl) {
       this._map.addControl(
@@ -110,7 +112,7 @@ class GrampsjsMap extends LitElement {
     }
     this._layercontrol = L.control.layers(
       {
-        OpenHistoricalMap: gl,
+        OpenHistoricalMap: this._gl,
         OpenStreetMap: tileLayer,
       },
       null,
@@ -118,12 +120,17 @@ class GrampsjsMap extends LitElement {
         position: 'bottomleft',
       }
     )
-    this._glMap = gl
     if (this.layerSwitcher) {
       this._map.addControl(this._layercontrol)
     }
+    this._map.on('baselayerchange', e => this._handleBaseLayerChange(e))
     this._map.invalidateSize(false)
     this._map.on('moveend', e => this._handleMoveEnd(e))
+  }
+
+  _handleBaseLayerChange(e) {
+    this._currentLayer = e.name
+    fireEvent(this, 'map:layerchange', {layer: this._currentLayer})
   }
 
   _handleMoveEnd(e) {
@@ -138,8 +145,12 @@ class GrampsjsMap extends LitElement {
   }
 
   updated(changed) {
-    if (changed.has('year') && this.year > 0) {
-      filterByDecimalYear(this._glMap._glMap, this.year)
+    if (
+      changed.has('year') &&
+      this.year > 0 &&
+      this._currentLayer === 'OpenHistoricalMap'
+    ) {
+      filterByDecimalYear(this._gl._glMap, this.year)
       return
     }
     if (this._map !== undefined) {
--- a/src/components/GrampsjsMap.js
+++ b/src/components/GrampsjsMap.js
@@ -18,7 +18,7 @@ const defaultConfig = {
   leafletZoomOffset: 0,
   glStyle: 'https://www.openhistoricalmap.org/map-styles/main/main.json',
   glAttribution:
-    '<a href="https://www.openhistoricalmap.org/copyright">OpenHistoricalMap</a>',
+    '<a href="https://www.openhistoricalmap.org/">OpenHistoricalMap</a>',
 }
 
 class GrampsjsMap extends LitElement {
--- a/src/components/GrampsjsMap.js
+++ b/src/components/GrampsjsMap.js
@@ -1,14 +1,13 @@
 import {html, LitElement} from 'lit'
-import {
-  Map,
-  TileLayer,
-  LatLng,
-  control,
-} from '../../node_modules/leaflet/dist/leaflet-src.esm.js'
+import 'leaflet'
+import 'maplibre-gl'
+import '@maplibre/maplibre-gl-leaflet'
 import './GrampsjsMapOverlay.js'
 import './GrampsjsMapMarker.js'
 import {fireEvent} from '../util.js'
 import '../LocateControl.js'
+
+const {L} = window
 
 const defaultConfig = {
   leafletTileUrl:
@@ -17,6 +16,7 @@ const defaultConfig = {
     '&copy; <a href="https://www.openstreetmap.org/copyright">OSM</a> contributors; <a href="https://carto.com/attributions">CARTO</a>',
   leafletTileSize: 256,
   leafletZoomOffset: 0,
+  glStyle: 'https://www.openhistoricalmap.org/map-styles/main/main.json',
 }
 
 class GrampsjsMap extends LitElement {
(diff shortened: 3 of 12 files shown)


src/components/GrampsjsMap.js
```diff
@@ -1,22 +1,22 @@
...
-} from '../../node_modules/leaflet/dist/leaflet-src.esm.js'
+import 'leaflet'
+import 'maplibre-gl'
+import '@maplibre/maplibre-gl-leaflet'
 import './GrampsjsMapOverlay.js'
 import './GrampsjsMapMarker.js'
 import {fireEvent} from '../util.js'
 import '../LocateControl.js'
 
+const {L} = window
```
1ec5, line 10: The MapLibre GL Leaflet plugin hooks into the `L` namespace that Leaflet exposes. This required refactoring many references to Leaflet in order to still build and run while satisfying ESLint. I’m not wedded to this approach; let me know if you can think of a less invasive way to build with these dependencies.

src/components/GrampsjsMap.js
```diff
@@ -1,22 +1,22 @@
...
+
 const defaultConfig = {
   leafletTileUrl:
     'https://{s}.basemaps.cartocdn.com/rastertiles/voyager/{z}/{x}/{y}{r}.png',
   leafletTileAttribution:
     '&copy; <a href="https://www.openstreetmap.org/copyright">OSM</a> contributors; <a href="https://carto.com/attributions">CARTO</a>',
   leafletTileSize: 256,
   leafletZoomOffset: 0,
+  glStyle: 'https://www.openhistoricalmap.org/map-styles/main/main.json',
```
1ec5, line 19: OHM offers [multiple official stylesheets](https://wiki.openstreetmap.org/wiki/OpenHistoricalMap/Reuse#Vector_tiles_and_stylesheets). For now, I’ve gone with the default “Historical” style. If desired, Gramps can define its own custom style, following the [MapLibre GL Style Specification](https://maplibre.org/maplibre-style-spec/). Tools like [Fresco](https://github.com/go-spatial/fresco/) and [Maputnik](https://github.com/maplibre/maputnik/) allow you to design custom styles visually.

src/components/GrampsjsFormEditLatLong.js
```diff
@@ -114,6 +114,7 @@ class GrampsjsFormEditLatLong extends GrampsjsObjectForm {
         <grampsjs-map
           latitude="${this.data.lat ? parseFloat(this.data.lat) : 0}"
           longitude="${this.data.long ? parseFloat(this.data.long) : 0}"
+          layerSwitcher
           mapid="edit-latlong-map"
```
1ec5, line 118: Some of these maps come with search boxes powered by OpenStreetMap’s Nominatim instance. OHM has its own Nominatim instance too. It’s a bit rough around the edges, but in the future, maybe we could switch to it whenever the map is showing the OHM layer, for consistency.

package.json
```diff
@@ -82,6 +82,7 @@
   "license": "AGPL",
   "dependencies": {
     "@hpcc-js/wasm": "^2.16.0",
+    "@maplibre/maplibre-gl-leaflet": "0.0.21",
```
1ec5, line 84: This compatibility shim works, but it preserves the Leaflet user experience, which is much less fluid than MapLibre GL JS. Migrating from Leaflet to MapLibre GL JS would enable smooth zooming, rotation, and better rendering performance. Something to consider for the future.

jeffreyameyer:
>  I’m unsure of where to get a relevant date,

Does this refer to "where to get the date from Gramps" or "where to get a date from OHM?". I believe it's Gramps, but I wasn't 100% sure.

1ec5:
I’m referring to getting dates within the current Gramps view that can be fed into the map component. For example, in one of the screenshots above, I’ve hovered over an event in 1895 – the map could automatically filter to 1895 while hovering over that item.

DavidMStraub:
Thanks so much, that looks awesome! 😍 I will review it as soon as I can, and for sure will be able to support on the dates question.

DavidMStraub:
Rebased to main.

DavidMStraub:
I finally had time to test it and it works & looks great! Also the changes to the code like the use of `L` make sense.

Concerning the time slider, I feel a more minimal slider would fit better - and we could use it regardless of map layer to only show pins of places with events in a given period.

Do I understand correctly the OHM map can either use a fixed date or a range?

1ec5:
Yes, the data is all available on the client side; it’s up to the client code to decide what to show or filter out. Unfortunately, the existing time slider library’s UI might be a bit much for some of the map views, but there would be nothing wrong with writing a homegrown UI just for Gramps.

With a slider, we’d need to decide the earliest and latest date the user should be allowed to choose. Alternatively, we could present the user with a date control. A range could simply entail showing two date controls instead of just one. Unfortunately, the HTML slider control only supports one knob, not two knobs, so we’d need to create and style a custom control if we need a double-knob slider to represent a date range.

1ec5:
Here’s what the map looks like with the new time slider from 1ec5/gramps-web#1:

Loveland in 1959

src/util.js
```diff
@@ -588,3 +588,120 @@ export function dateIsEmpty(date) {
...
+      newFilter[2][2][2] = decimalYear
+    }
+    return newFilter
+  }
+
+  const dateFilter = [
+    'all',
+    ['any', ['!has', 'start_decdate'], ['<', 'start_decdate', decimalYear + 1]],
+    ['any', ['!has', 'end_decdate'], ['>=', 'end_decdate', decimalYear]],
+  ]
```
1ec5, line 669: This uses the [legacy filter syntax](https://maplibre.org/maplibre-style-spec/deprecations/#function) from the MapLibre Style Specification. Eventually, once the Historical style is migrated to the newer expression syntax in OpenHistoricalMap/issues#775, we’ll need to update this code accordingly. It’ll be quite similar, something like this:

```js
  const dateFilter = [
    'all',
    ['any', ['!', ['get', 'start_decdate']], ['<', ['get', 'start_decdate'], decimalYear + 1]],
    ['any', ['!', ['get', 'end_decdate']], ['>=', ['get', 'end_decdate'], decimalYear]],
  ]
```

But we’ll probably want to take advantage of the `let` expression to simplify this method overall:

```js
  const dateFilter = [
    'let',
    'decimalYear', decimalYear,
    [
      'all',
      ['any', ['!', ['get', 'start_decdate']], ['<', ['get', 'start_decdate'], ['+', ['var', 'decimalYear'], 1]]],
      ['any', ['!', ['get', 'end_decdate']], ['>=', ['get', 'end_decdate'], ['var', 'decimalYear']]],
    ],
  ]
```

A bit more verbose, but then on subsequent applications, we’ll be able to manipulate the filter less invasively and probably more performantly:

```js
  if (filter && filter[0] == 'let' && filter[1] == 'decimalYear') {
    newFilter[2] = decimalYear
    return newFilter
  }
```

DavidMStraub:
I now added filtering the map pins by the time slider.

I also added two additional buttons:

- the cogwheel allows to change the "tolerance" (central year &pm; x) for place filtering
- the switch allows to disable/enable place filtering

grafik

In the OSM view, the slider is disabled when the switch is off, while in the OHM view the slider is always enabled as it not only filters places but also the map.

I have three open points:

- [ ] Calendars other than Gregorian are not handled properly yet
- [ ] The layer control should be on top of the time slider, not below
- [ ] The OHM layer is initially not filtered

I don't know how to fix the last point. Applying the filter after switching layers does not work as the map is not fully loaded yet. And I haven't managed to find an appropriate event to listen to. :face_with_diagonal_mouth:

src/components/GrampsjsMap.js
```diff
@@ -67,66 +70,94 @@ class GrampsjsMap extends LitElement {
...
+      },
+      null,
+      {
+        position: 'bottomleft',
+      }
+    )
     if (this.layerSwitcher) {
       this._map.addControl(this._layercontrol)
     }
+    this._map.on('baselayerchange', e => this._handleBaseLayerChange(e))
```
1ec5, line 129: > The OHM layer is initially not filtered

Add an event listener for the [`styledata`](https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/MapEventType/#styledata) event. This will fire after MapLibre has downloaded enough of the stylesheet to manipulate it, but before the tiles are done loading:

```js
const mapLibreMap = this._gl.getMaplibreMap()
mapLibreMap.on('styledata', e => filterByDecimalYear(mapLibreMap, this.year))
```

src/components/GrampsjsMap.js
```diff
@@ -67,66 +70,94 @@ class GrampsjsMap extends LitElement {
...
   }
 
-  updated() {
+  updated(changed) {
+    if (
+      changed.has('year') &&
+      this.year > 0 &&
+      this._currentLayer === 'OpenHistoricalMap'
+    ) {
+      filterByDecimalYear(this._gl._glMap, this.year)
```
1ec5: `this._gl.getMaplibreMap()` is probably more robust than accessing this internal property.

1ec5 (commented):
> Calendars other than Gregorian are not handled properly yet

How does Gramps handle non-Gregorian dates in general? Do they get normalized to Gregorian? Is there a conversion utility for situations where there isn’t a risk of dataloss?

On the OHM side, recording non-Gregorian dates is still an open question, but I suspect the tiles will ultimately normalize them to the proleptic Gregorian calendar for a consistent display.

DavidMStraub:
> How does Gramps handle non-Gregorian dates in general? Do they get normalized to Gregorian? Is there a conversion utility for situations where there isn’t a risk of dataloss?

Dates are stored as [year, month, day] and calendar type (Gregorian, Julian, Hebrew, Islamic, ...). Conversions are done by implementing functions [year, month, day] &harr; numeric Julian day number for each calendar.

I could change something in the backend to use the existing Python functions for the conversion; the alternative would be to implement the functions in [this module](https://github.com/gramps-project/gramps/blob/master/gramps/gen/lib/gcalendar.py) in Javascript.

For the time being I wouldn't even feel too bad treating Julian dates as Gregorian and ignoring all other dates.

src/components/GrampsjsMap.js
DavidMStraub: Yes, thanks!! I now put it into the `baselayerchanged` event handler, because apparently that `styledata` event listener has to be recreated each time the layer is changed, but now it works!

DavidMStraub:
To solve the overlap issue I now simply moved the slider below the map. That makes it wider which is helpful on mobile (especially in portrait mode).

grafik

src/components/GrampsjsMap.js
```diff
@@ -17,6 +17,8 @@ const defaultConfig = {
   leafletTileSize: 256,
   leafletZoomOffset: 0,
   glStyle: 'https://www.openhistoricalmap.org/map-styles/main/main.json',
+  glAttribution:
+    '<a href="https://www.openhistoricalmap.org/copyright">OpenHistoricalMap</a>',
```
1ec5: OHM data is in the public domain (except where otherwise noted), so this attribution isn’t required, but the project appreciates it as a courtesy. In this case, I figure it would help to keep the Gramps forum from getting confused questions about why this or that isn’t available in the OHM layer. 🙂

1ec5 (commented):
> For the time being I wouldn't even feel too bad treating Julian dates as Gregorian and ignoring all other dates.

That sounds pretty reasonable to me for the time being, since the default tolerance is already much more generous than the Julian offset for a given date. Since countries adopted the Gregorian calendar in different years, we’d be showing pins from different calendars simultaneously. That probably calls for the time slider to have its own calendar option, eventually.

src/components/GrampsjsMap.js
DavidMStraub: Yes, makes sense!

DavidMStraub:
Handling all calendars turned to be much simpler than I though, if we accept &pm;1 inaccuracy Swedish, Julian, and Gregorian are all the same; French, Hebrew and Persian are just a constant offset; and Islamic is a linear function.

I also made the slider's maximum extent set automatically to the earliest event in the tree rather than hard-coding it to 1500.

Now I want to add tooltips to the 2 new buttons and we're done I think :slightly_smiling_face:

src/views/GrampsjsViewMap.js
```diff
@@ -389,12 +392,25 @@ export class GrampsjsViewMap extends GrampsjsView {
...
 
+  _getMinYear() {
+    const years = this._dataEvents
+      ?.filter(event => event.place)
+      ?.map(event => getGregorianYears(event.date)?.[0])
+      ?.filter(y => y !== undefined)
+    let minYear = Math.min(...years)
+    const lastYear = new Date().getFullYear() - 1
+    minYear = Math.min(minYear, lastYear)
+    minYear = Math.max(minYear, 1) // disallow negative
```
1ec5, line 410: By the way, the map data and filtering code support BCE dates. Quite unlikely, of course, but I mention this in case we do decide to allow it in the future.

The only gotcha is that the year is offset by one because there’s no year 0, per ISO 8601. Currently this component is using −1 as a magic number for an unset year; we’d need to use something else like `undefined`.

1ec5 (commented):
The calendar conversions here are good motivation for OHM to add formal support for non-Gregorian dates. I think we’ll try to do it in a way that’s backwards-compatible with the functionality we’re using here, but I’ll keep you posted if anything needs tweaking later.

src/views/GrampsjsViewMap.js
DavidMStraub: Yes you're right. I think it's pretty ok for Gramps Web, but of course not ok for OHM in general. So would be more elegant to handle in full generality in the future here as well.

PR #479, opened 2024-08-07
Replace custom date filtering with maplibre-gl-dates

1ec5:
This replaces the custom date filtering code for the OpenHistoricalMap basemap with [maplibre-gl-dates](https://github.com/OpenHistoricalMap/maplibre-gl-dates/), which is forward-compatible with expression-based filters.


1ec5:
I’m unable to test this change locally because of the following error:

```
gramps-frontend-dev  | Error while transforming src/components/GrampsjsMap.js: Could not resolve import "@openhistoricalmap/maplibre-gl-dates".
gramps-frontend-dev  | 
gramps-frontend-dev  |   3 | import 'maplibre-gl'
gramps-frontend-dev  |   4 | import '@maplibre/maplibre-gl-leaflet'
gramps-frontend-dev  | > 5 | import '@openhistoricalmap/maplibre-gl-dates'
gramps-frontend-dev  |     |        ^
gramps-frontend-dev  |   6 | import './GrampsjsMapOverlay.js'
gramps-frontend-dev  |   7 | import './GrampsjsMapMarker.js'
gramps-frontend-dev  |   8 | import {fireEvent} from '../util.js'
gramps-frontend-dev  | 
```

I’m not sure why this error occurs, since the `@openhistoricalmap/maplibre-gl-dates` package has been installed. One unusual aspect is that, as long as this comes after the `maplibre-gl` import, the module doesn’t import anything; it only adds a method to an existing global `maplibregl` object.

DavidMStraub:
Thanks, big simplification!

I got the import error at first as well, but it worked after restarting the web dev server (`npm run start`).

It works as expected in the browser, but I see a lot of errors in the console, they are all of the type

`layers.land.filter[3][0]: expected one of [==, !=, >, >=, <, <=, in, !in, all, any, none, has, !has], "let" found`

It looks like this error occurs for all layers.

1ec5:
Thanks, I fixed the error in OpenHistoricalMap/maplibre-gl-dates@d6b6da1a8314bb8758021074e7db45669c8dfa4b and upgraded the library.

PR #522, opened 2024-11-01
Ignore index.html in service worker cache

MrTrustor:
Fixes #517

By default, index.html is cached by the service worker. This breaks the application if it's hosted behind an authentifying proxy which requires periodic revalidation of an auth token. Specifically, once the auth token expires:
1. The index.html is still served from the service worker cache.
2. Subsequent requests (e.g for JS resources) are redirected to the login page of the proxy.
3. That login page is never displayed to the user as index.html has been successfully loaded.
4. All requests but index.html fail, and the app displays a blank page.

This change fixes this behavior by preventing index.html from being cached by the service worker. With that change, once the auth token from the proxy expires:
1. The request for index.html is not served from the service worker cache.
2. Instead, a network request for index.html is made, and the user is properly redirected to the proxy login page.
3. The user logs in, and index.html is served as normal.
4. Subsequent requests (e.g for JS resources) succeed as now the auth token has been refreshed.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -12,6 +12,7 @@ const baseConfig = createSpaConfig({
   developmentMode: process.env.ROLLUP_WATCH === 'true',
   injectServiceWorker: true,
   workbox: {
+    globIgnores: ['index.html'],
     navigateFallbackDenylist: [/^\/api.*/],
     skipWaiting: false,
     clientsClaim: false,


DavidMStraub:
This shouldn't affect users that do *not* use an authentifying proxy, right?

On my setup, I've recently noticed I sometimes get a blank page on mobile when I open the app, I wonder whether it's related. This started a few months ago.

MrTrustor:
I don't think this would have any impact on users not using a proxy (the requests just happen without any redirection from a proxy since there is none) but it might be worth double checking.

On Safari on iOS, I've always run into this blank page issue, which I'm assuming is the same I see on desktop. Though, if you're not using a proxy, I don't know what could be causing it.

MrTrustor:
@DavidMStraub Are you waiting for something from me on this PR?

DavidMStraub:
Nope :slightly_smiling_face:

Issue #521, opened 2024-10-28
Thumbnails in Relationship Graph don't all display


jx24734:
**Install Method**
Docker-Compose

**System information**
Gramps 5.2.2
Gramps Web API 2.4.2
Gramps Web Frontend 24.8.0
Gramps QL 0.3.0
locale: en
multi-tree: false
task queue: true

**Issue**
When the Relationship Graph is expanded beyond 3 degrees of separation, not all thumbnails display and ones that were previously displaying disappear, and the person's name shifts to the left where the thumbnail should be.
 
Relationship Graph view
![image](https://github.com/user-attachments/assets/57a39a41-581a-4f99-adf5-8f9c524228a5)

Person Details
![image](https://github.com/user-attachments/assets/88886922-fe2f-4baa-a729-0061b4b1f544)

DavidMStraub:
Implemented in #524 :rocket:

PR #524, opened 2024-11-04
RelationshipChart: make maxImages configurable

geostag:
make maximum number  of images displayed in relationship graph UI-configurable. solves #521


DavidMStraub:
Perfect, and works like a charm!

Issue #376, opened 2024-02-01
Add tree ID, username and permissions level to System Information


emyoulation:
Some complaints will be because the Permission Level doesn't enable the feature a user reads about. (And sometimes the "Owner" will forget which tabs are accessible at lower levels.) The User won't be able to tell us about their Permission Level because only users at Owner level have access to the "Manage User" preferences tab can see the Permissions Level of any particular account.  

We could spend a lot of energy on Discourse troubleshooting a simple Permissions issue for no reason. 

* The Username and permissions level should show beside the "LOGOUT" button in User Settings 
* The tree ID, username and permissions level should be included in the System Information that can be "Copy to clipboard"

DavidMStraub:
Agreed, except tree ID & username should not be in the copied data as they can represent personally identifiable information.

PR #527, opened 2024-11-16
🩹 Fixing static app-title

letsjustfixit:
Fixing: that the AppBar has no access to the _dbInfo property so the app-title stays always static.


DavidMStraub:
That's true. Funnily enough, it's a leftover from the ancient Gramps Web app frontend (pre Gramps Web API), and I never bothered to change it (i.e. make it display the database name again) because I don't think it makes sense right now as long as we don't add the possibility to change the database name from the frontend. Many trees have names that are not really suitable as app header. So the to do list is

- make database name changeable in Web API - this is already done, although I don't remember if it works only in the multi tree setup
- make database name changeable in the frontend
- leave one release (1 month) between this and changing the app header, to leave tree owners time to change

letsjustfixit:
Got it. 

(For me this was essential because I have multiple trees served by the same instance and they even have some common people. 
I was even thinking about assigning different color schemes to trees to make it more obvious when working with the different trees in different browser windows at the same time. I've also added the current logged in username to the user settings screen above the logout to make it obvious which test user I'm logged in with. )

We could add an optional TREE_DISPLAY_NAME property that defaults to the existing tree name as TREE is an essential reference key and TREE_DISPLAY_NAME might contain special characters illedavised to use for a key / folder name(accented characters,, unicode special characters etc..) .

DavidMStraub:
> I've also added the current logged in username to the user settings screen above the logout to make it obvious which test user I'm logged in with. 

Yes, #376, urgently needed.

DavidMStraub:
Unblocked! However, I'd wait one release before implementing this, so users have time to update their backend configuration to allow renaming their tree, before it becomes the default title.

DavidMStraub:
Rebased/fixed. Will merge after next release!

PR #532, opened 2024-11-23
Add user information to settings pages

### LINKED ISSUE #376: Add tree ID, username and permissions level to System Information

letsjustfixit:
closes #376 
Extended SystemInfo
Extended Logout

- [X]     The Username and permissions level should show beside the "LOGOUT" button in User Settings
- [X]     The tree ID, username and permissions level should be included in the System Information that can be "Copy to clipboard"
- [X]     tree ID & username should not be in the copied data as they can represent personally identifiable information.

### DIFF
--- a/src/components/GrampsjsSysinfo.js
+++ b/src/components/GrampsjsSysinfo.js
@@ -31,18 +31,20 @@ export class GrampsjsSysinfo extends GrampsjsTranslateMixin(LitElement) {
   static get properties() {
     return {
       data: {type: Object},
+      tree: {type: String},
     }
   }
 
   constructor() {
     super()
     this.data = {}
+    this.tree = ''
   }
 
   render() {
     const version = '[VI]{version}[/VI]'
     return html`
-      <div class="copy" id="copy">
+      <div class="copy">
         <mwc-icon-button
           icon="content_copy"
           id="copy-button"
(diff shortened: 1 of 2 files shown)


DavidMStraub:
Nice, thanks!

For the sysinfo tab, the UI is misleading though - the gray clipboard icon with the gray rounded box clearly communicates to the user that they are copying this box's content. Also, the user name and role are not part of the information about the system. So my proposal is:

- remove user name and role from the sys info tab
- Move the tree ID below the gray rounded box

Concerning the user name and role, in my opinion `admin [Administrator]` is not clear enough. Let's be explicit about what they represent (user name and role).

letsjustfixit:
How about this?
Tree below
User-role

Issue #27, opened 2021-08-07
Show alternate place names in place view


DavidMStraub:

cajturner:
This is done 🙌

Ziesie1:
@cajturner what do you think about using ther alternate place names for time-releated events. If a person was born in 1930 and a place have an alternative name between 1910 and 1945, it would be great if the user could see the alternate place name. Currently, the main place name is always displayed. What do you think? (Perhaps this should be a separate feature)

DavidMStraub:
Gramps Desktop can do that, so yes, it would be nice, but it would be a separate feature and one for the backend first (event profile).

PR #566, opened 2025-01-05
Display alternative place names

cajturner:
This is my first contribution and an attempt to solve issue: https://github.com/gramps-project/gramps-web/issues/27

I have decided to only tackle displaying alternative place names and not implementing edit functionality. This will be added in a future PR.

Any feedback or suggestions welcome

## Screenshots
Place view when alternative name is present
Screenshot 2025-01-05 at 20 53 26
Alternative name Tab when alternative name is present
Screenshot 2025-01-05 at 20 53 28
Alternative name Tab when alternative name is present and in edit mode
Screenshot 2025-01-05 at 20 53 31
Place view when alternative name is not present
Screenshot 2025-01-05 at 20 53 37
Place view when alternative name is not present and in edit mode
Screenshot 2025-01-05 at 20 53 40

### DIFF
--- a/src/components/GrampsjsObject.js
+++ b/src/components/GrampsjsObject.js
@@ -22,6 +22,7 @@ import './GrampsjsIcon.js'
 import './GrampsjsNames.js'
 import './GrampsjsPlaceChildren.js'
 import './GrampsjsPlaceRefs.js'
+import './GrampsjsPlaceNames.js'
 import './GrampsjsGallery.js'
 import './GrampsjsMap.js'
 import './GrampsjsMapMarker.js'
@@ -56,6 +57,11 @@ const _allTabs = {
       data.placeref_list?.length > 0 ||
       ('placeref_list' in data && data?.backlinks?.place?.length >= 0),
     conditionEdit: data => 'placeref_list' in data,
+  },
+  placeNames: {
+    title: 'Alternate Names',
+    condition: data => data.alt_names?.length > 0,
+    conditionEdit: data => data.alt_names?.length > 0,
   },
   map: {
     title: 'Map',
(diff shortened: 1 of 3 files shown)


src/components/GrampsjsObject.js
```diff
@@ -59,6 +60,11 @@ const _allTabs = {
       ('placeref_list' in data && data?.backlinks?.place?.length >= 0),
     conditionEdit: data => 'placeref_list' in data,
   },
+  placeNames: {
+    title: 'Alternate Names',
+    condition: data => data.alt_names?.length > 0,
```
DavidMStraub: Probably it doesn't matter in practice, but since this function is also called for objects that are not places and thus don't have an `alt_names` property, I find it clearer to add the `?.` also in front of `alt_names`

cajturner:
> Can you add the language as well?
I added the 'Alternate Names' string to strings.js and can see the value translated correctly. Is anything else required?
Screenshot 2025-01-06 at 20 17 55


> The main issue I see is with date formatting - unfortunately we don't have a proper way in Javascript to represent all of the complexity of Gramps' date formatting, and just using `Y-M-D` can be misleading (already shown in your example with zeros). This is particularly important as place name dates are likely to always be spans.
> 
> I'm afraid the only proper solution would be to add the formatted date to the place profile in the backend first, here: https://github.com/gramps-project/gramps-web-api/blob/master/gramps_webapi/api/resources/util.py#L362
> 
> A more radical alternate solution would be to add a new date formatting endpoint in the API where the frontend can post a date object and it returns a string...
I'm happy to add the formatted date to the place profile backend. I can see `locale.date_displayer.display()` being used so that looks like it can be reused

DavidMStraub:
> I'm happy to add the formatted date to the place profile backend. I can see locale.date_displayer.display() being used so that looks like it can be reused

Perfect, let me know if you need guidance.

cajturner:
Thanks @DavidMStraub I've been busy since last week but hope to still make the changes to the api. I was struggling a little to get the tests to run on my local machine (mac). I'll give it another go over the weekend

DavidMStraub:
The backend development docs haven't been updated in a while, if anything doesn't work as expected when setting up the dev environment, feel free to open an issue about it.

DavidMStraub:
Converted to draft as we first need to finish https://github.com/gramps-project/gramps-web-api/pull/608.

cajturner:
I'll pick this up again as the api changes have been merged

DavidMStraub:
Hi, FYI I intend to release v25.6.0 this weekend, let me know if you'd like to merge this before.

cajturner:
image
I can see the alternative place name now being shown. 

@DavidMStraub Do we need anything else before this can be merged?

DavidMStraub:
Thanks for rebasing, looks good!

PR #568, opened 2025-01-11
Move DNA matches & chromosome browser to separate DNA view

### LINKED ISSUE #545: Create separate DNA view
- Add a DNA entry to the main menu (we can use `mdiDna` as icon)
- Initially populate it with the existing chromosome browser view. We can use a dropdown to select people with existing match data.

DavidMStraub:
Fixes #545.

To do before this can be merged: 

- ~~Add sort buttons to new responsive table~~ descoped
- [x] Remove DNA view from person page
- [x] Allow deleting matches
- [x] Allow editing matches
- [x] Replace P/M by paternal/maternal


DavidMStraub:
Depends on https://github.com/gramps-project/gramps-web-api/pull/604.

DavidMStraub:
Need some more error & edge case handling & some polishing, otherwise done.

Issue #227, opened 2023-06-14
can't easily create parent


socphoenix:
On the gramps desktop interface it automatically pops up a blank spot you can double click to add a parent.  On the web-app this doesn't happen, and you can't add parents through person details. Looks like I have to add a person, then find the child in the tree, and link the parent. It would be much more convenient to be able to edit these details from the person detail view.

cajturner:
Happy to support this work, where I can. @DavidMStraub Let me know if you have a direction in mind and any task I could contribute to

DavidMStraub:
Sure! 

I think the first step would be to create a new pop-up dialog for creating a new person. At the moment, we have two different ways of adding new objects: via a dedicated page/view or via a pop-up. To not duplicate code, both elements share a mixin. See here for events:

https://github.com/gramps-project/gramps-web/blob/main/src/views/GrampsjsViewNewPerson.js
https://github.com/gramps-project/gramps-web/blob/main/src/components/GrampsjsFormNewEvent.js
https://github.com/gramps-project/gramps-web/blob/main/src/mixins/GrampsjsNewEventMixin.js

For people, only the first one exists an the mixin has not been separated out yet. Once that is done, adding the form to the person view is similar to the "new event" form, which also allows creating an event. The logic for handling the completed & submitted form is in `handleEditAction` of https://github.com/gramps-project/gramps-web/blob/main/src/views/GrampsjsViewObject.js.

cajturner:
I have opened a draft PR where I create a newPersonMixin. https://github.com/gramps-project/gramps-web/pull/575

Question about the next steps, should we allow a new person to be created within another dialog modal? For example, in the screenshot below should we add a + button next to father that opens a new dialog or should it expand the existing dialog. 

![Image](https://github.com/user-attachments/assets/155bd58a-ef06-4b88-8aa3-c045221cbab8)

DavidMStraub:
Stacking models on top of each other is indeed something I would avoid. But maybe we can make the UI simpler than having a a big dialog with expanding sections. Instead of a single dialog to edit father, mother and relationship at the same time, we could have a single dialog to edit the relationship, one dialog for editing father and one for editing mother.  For editing a parent, we could either have a single dialog where the user can switching between selecting an existing and adding a new person (by unhiding a NewPersonMixing form on a button click), or we could have two separate icon buttons in the UI for "link existing person as father/mother" and "create new person as father/mother", then the individual popups are even simpler. What do you think?

[...]

PR #575, opened 2025-01-20
Prep for new Person dialog

cajturner:
https://github.com/gramps-project/gramps-web/issues/227

Move form to mixin in prep for new Person dialog


DavidMStraub:
Do you think it makes sense rebasing this or should we rather start over, given the changes to `GrmapsjsViewPerson` in the meantime?

cajturner:
> Do you think it makes sense rebasing this or should we rather start over, given the changes to `GrmapsjsViewPerson` in the meantime?

Let's close and start over. Hopefully will have some more time to be able to contribute again in the not so distant future

Issue #304, opened 2023-11-05
Allow deleting users


DavidMStraub:
It's currently not possible to delete users from the frontend, but the endpoint exists.

DavidMStraub:
Fixed by #577!

PR #577, opened 2025-01-23
Add delete user button

DanielMorsch:
Hi,

this is my first contribution and also my first steps in JS.

I added the delete user button to solve issue #304 
I have oriented myself to the edit button and the other dialogs in the users view. Happily, the API was already available.
 
Feedback is always appreciated 😄 

### Screenshots
Delete button inside table
Bildschirmfoto 2025-01-23 um 16 21 27
 Confirmation dialog
Bildschirmfoto 2025-01-23 um 16 21 38

### Open things
- Add Language strings somewhere to translate into other languages

### DIFF
--- a/src/SharedStyles.js
+++ b/src/SharedStyles.js
@@ -264,6 +264,10 @@ export const sharedStyles = css`
     --mdc-theme-on-primary: var(--mdc-theme-on-secondary);
   }
 
+  .delete {
+    color: #bf360c;
+  }
+
   mwc-icon-button.large {
     --mdc-icon-size: 32px;
   }
--- a/src/SharedStyles.js
+++ b/src/SharedStyles.js
@@ -264,10 +264,6 @@ export const sharedStyles = css`
     --mdc-theme-on-primary: var(--mdc-theme-on-secondary);
   }
 
-  .delete {
-    color: #bf360c;
-  }
-
   mwc-icon-button.large {
     --mdc-icon-size: 32px;
   }
--- a/src/components/GrampsjsUsers.js
+++ b/src/components/GrampsjsUsers.js
@@ -41,7 +41,6 @@ export class GrampsjsUsers extends GrampsjsTableBase {
           <th>${this._('E-mail')}</th>
           <th>${this._('Role')}</th>
           <th></th>
-          <th></th>
         </tr>
         ${this.data.map(
           (obj, index) => html`
@@ -60,10 +59,8 @@ export class GrampsjsUsers extends GrampsjsTableBase {
                 <grampsjs-tooltip for="button-edit-${index}">
                   ${this._('Edit user')}
                 </grampsjs-tooltip>
-              </td>
-              <td>
                 <mwc-icon-button
-                  class="delete"
+                  class="error"
                   icon="delete_forever"
                   @click="${e => this._handleDeleteClick(e, obj.name)}"
                   id="button-del-${index}"
@@ -265,18 +262,18 @@ export class GrampsjsUsers extends GrampsjsTableBase {
   }
 
   _deleteUserDialog(username) {
-    const [user] = this.data.filter(el => el.name === username)
     return html`
       <mwc-dialog
         scrimClickAction=""
         escapeKeyAction=""
         open
       >
-        <div>${this._('Delete this user?')}</div>
+        <div>${this._('Do you really want to delete user \"%s\"?', username)}</div>
+        <div>${this._('This action cannot be undone.')}</div>
         <mwc-button
           slot="primaryAction"
           dialogAction="delete"
-          @click="${e => this._handleDelete(username)}">
+          @click="${() => this._handleDelete(username)}">
           ${this._('_Delete')}
         </mwc-button>
         <mwc-button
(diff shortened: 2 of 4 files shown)


dstraub-pf:
This is beautiful! Thanks for the screenshot as well!

Perhaps we can make the dialog a bit more verbose, something like "Do you really want to delete user %s? This action cannot be undone." (since deleting Gramps objects can actually be undone.)

To make it translatable, simple add the exact string to the end of `en.json`, everything else will happen automagically thanks to Weblate!

DavidMStraub:
This is beautiful! Thanks for the screenshot as well!

Perhaps we can make the dialog a bit more verbose, something like "Do you really want to delete user %s? This action cannot be undone." (since deleting Gramps objects can actually be undone.)

To make it translatable, simple add the exact string to the end of en.json, everything else will happen automagically thanks to Weblate!

src/SharedStyles.js
```diff
@@ -264,6 +264,10 @@ export const sharedStyles = css`
     --mdc-theme-on-primary: var(--mdc-theme-on-secondary);
   }
 
+  .delete {
+    color: #bf360c;
```
DavidMStraub: There is a class `.error` in [SharedStyles](https://github.com/gramps-project/gramps-web/blob/main/src/SharedStyles.js) that you can use instead of `.delete`, then no need to specify it here.

src/components/GrampsjsUsers.js
```diff
@@ -41,6 +41,7 @@ export class GrampsjsUsers extends GrampsjsTableBase {
           <th>${this._('E-mail')}</th>
           <th>${this._('Role')}</th>
           <th></th>
+          <th></th>
```
DavidMStraub: What about putting it in the same column as the edit button? Right now they look a bit far apart (nitpicking...).

src/components/GrampsjsUsers.js
```diff
@@ -60,6 +61,17 @@ export class GrampsjsUsers extends GrampsjsTableBase {
...
               </td>
+              <td>
+                <mwc-icon-button
+                  class="delete"
+                  icon="delete_forever"
+                  @click="${e => this._handleDeleteClick(e, obj.name)}"
+                  id="button-del-${index}"
+                ></mwc-icon-button>
+                <grampsjs-tooltip for="button-del-${index}">
+                  ${this._('Delete user')}
```
DavidMStraub, line 69: Also this string needs to be added to `en.json`

src/components/GrampsjsUsers.js
```diff
@@ -246,6 +264,30 @@ export class GrampsjsUsers extends GrampsjsTableBase {
     `
   }
 
+  _deleteUserDialog(username) {
+    const [user] = this.data.filter(el => el.name === username)
```
DavidMStraub: This line is not needed

src/components/GrampsjsUsers.js
```diff
@@ -246,6 +264,30 @@ export class GrampsjsUsers extends GrampsjsTableBase {
...
+      <mwc-dialog
+        scrimClickAction=""
+        escapeKeyAction=""
+        open
+      >
+        <div>${this._('Delete this user?')}</div>
+        <mwc-button
+          slot="primaryAction"
+          dialogAction="delete"
+          @click="${e => this._handleDelete(username)}">
```
DavidMStraub: `() => this._handleDelete(username)`

src/views/GrampsjsViewUserManagement.js
```diff
@@ -109,6 +110,11 @@ export class GrampsjsViewUserManagement extends GrampsjsView {
     })
   }
 
+  _handleUserDeleted(e) {
+    const data = e.defail
```
DavidMStraub: Typo, but anyway this is not used

src/SharedStyles.js
DanielMorsch: Okay great. I will use the `.error` class.

src/components/GrampsjsUsers.js
DanielMorsch: great. thats easy.

src/components/GrampsjsUsers.js
DanielMorsch: Yeah. In my first try i added it in the same column. 
in a big resolution it looks good.
Bildschirmfoto 2025-01-23 um 17 10 36
but if you open it in lower resolution like phone it swaps to a vertical layout:
Bildschirmfoto 2025-01-23 um 17 13 59
that stopped me from putting it in the same column. 
what is your opinion?

DavidMStraub: I like it actually :slightly_smiling_face: 

Besides, part of the reason I thought about this is that I'm working on more responsive tables in a separate PR, which wrap from columns to individual cells on small screens, and in that case it would be better if they are in the same cell.

DanielMorsch:
> Perhaps we can make the dialog a bit more verbose, something like "Do you really want to delete user %s? This action cannot be undone." (since deleting Gramps objects can actually be undone.)

Sure. Thats better. Like this one?
Bildschirmfoto 2025-01-23 um 17 26 26

src/components/GrampsjsUsers.js
DanielMorsch: okay 👌

DanielMorsch:
All tests should now be successful.

PR #600, opened 2025-02-11
Use d3 links for connectors in relationship chart

DavidMStraub:
Hi @geostag, I finally got around to replacing the Graphviz connectors in the relationship chart with smooth d3 ones, what do you think?

I grew so fond of the relationship chart that I started thinking we should use the same function also for ancestor/descendant/hourglass - they could just correspond to different data filters rather than chart types. But having something that is left-to-right rather than top-to-bottom would still be nice.

grafik

### DIFF
--- a/src/charts/RelationshipChart.js
+++ b/src/charts/RelationshipChart.js
@@ -1,5 +1,6 @@
 import {create, select} from 'd3-selection'
 import {zoom} from 'd3-zoom'
+import {linkVertical} from 'd3-shape'
 import {Graphviz} from '@hpcc-js/wasm'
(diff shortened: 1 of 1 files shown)


geostag:
very nice, much better than the original graphviz connectors, i like that!
Using only the relationship chart would have the disadvantage of not being able to control order in one level. E.g.  control, that females occur always on the same branch in the ancestor chart. To address this, you would have to trick graphviz. Therefor I suppose to keep the current charts. However, creating a left to right chart would be also possible using the underlying graphviz in relationship chart.

Issue #336, opened 2023-12-26
Confirm password in registration form


DavidMStraub:
The registration form should have two password fields to reduce the likelihood of mistyping the password.

DavidMStraub:
No 🙂

DanielMorsch:
Okay, had just played around a little.

<img width="397" alt="Image" src="https://github.com/user-attachments/assets/427de6f8-b4db-4566-9de6-3deb4912f641" />

But had to use the `md-outlined-text-field` instead of the old deprecated `mwc-textfield` ones.

DavidMStraub:
> It was planned to replace them at some point anyway, right?

Yes indeed!

PR #612, opened 2025-02-23
Confirm password in registration form

DanielMorsch:
Hi,

take some progress on #336


DavidMStraub:
Thanks!

I made your life a bit harder because I merged a PR this week changing a couple of things. You will have to replace `GrampsjsTranslateMixin` by `GrampsAppStateMixin`, and all places where strings are passed (`.strings="${this.strings}"`) now take `appState` instead (`.appState="${this.appState}"`) . Let me know if there are any other conflicts, I can help fix them.

DanielMorsch:
No problem. Was not that complicated.

DanielMorsch:
@DavidMStraub 
i merge your changes into my branch but cant open the web anymore. got some errors to refresh 
Bildschirmfoto 2025-02-23 um 20 20 15
is this only happening to me or also on your site if you don't have a refresh token yet? 😅

DavidMStraub:
Not sure how this happened, but maybe due to an inconsistent state during rebasing.  Please go to Local Storage in the browser dev tools and delete the access token and refresh token keys, it will allow you to login again.

DanielMorsch:
There are no access token and refresh tokens saved on my side.
it is not clear to me why this is a mistake at all. 
I want to call up the login page and at this point I don't even have a token.

DanielMorsch:
Bildschirmfoto 2025-02-25 um 17 36 40

i do not see even the login page because the error happened in loading the page.

DavidMStraub:
Ok got it, that's indeed a regression of the PR I merged. Will fix.

DavidMStraub:
Should be fixed in main.

DanielMorsch:
@DavidMStraub 
Ready to review 😀

src/components/GrampsjsFormRegister.js
```diff
@@ -0,0 +1,250 @@
...
+              label="${this._('Password')}"
+              @input="${this._handlePasswordInput}"
+              type="password"
+            ></md-outlined-text-field>
+            <md-outlined-text-field
+              required
+              outlined
+              autocapitalize="off"
+              id="password2"
+              error-text="${this._('Passwords do not match.')}"
```
DavidMStraub, line 107: I think this is missing in en.json?

DavidMStraub:
Looks good! Can you please check that all strings are contained in `en.json` or `strings.js`? I think there are a couple of them missing.

src/components/GrampsjsFormRegister.js
DanielMorsch: its already in en.json

DanielMorsch:
I have found another missing string. but they should all be there now.

DavidMStraub:
Thanks, one final string request: can you please change "Login to" to "Log in to"? https://english.stackexchange.com/questions/5302/log-in-to-or-log-into-or-login-to

One thing that's a bit odd is that when clicling on the register link and then again on the login link, the browser back buttons don't work correctly. I think it is because of the `pushState` that is done in `GrampsJs.js`.

DanielMorsch:
> Thanks, one final string request: can you please change "Login to" to "Log in to"? https://english.stackexchange.com/questions/5302/log-in-to-or-log-into-or-login-to

Thats fine for me. I changed it to "Log in to".

> One thing that's a bit odd is that when clicling on the register link and then again on the login link, the browser back buttons don't work correctly. I think it is because of the `pushState` that is done in `GrampsJs.js`.

Good catch. I think we do not need to handle the pushState manually anymore. Works fine for me without. I assume that the things were only in there because both urls were for the same component with different parameters. Or is there another use case that I am not looking at?

DavidMStraub:
Great, ready to merge! Can you please rebase? Apparently there are merge conflicts.

DavidMStraub:
Rebased.

HxgnDaniel:
Thank you for rebasing.

DanielMorsch:
Thank you for rebasing.

Issue #444, opened 2024-05-22
Please add support for custom name display formats


syominsergey:
Hi!

In my culture preferrable order of names looks like this:
```
<given name> <patronymic surname> <surname>
```
Or another order also possible:
```
<surname> <given name> <patronymic surname>
```

But in GrampsWeb another order of names is widely used:
```
<given name> <surname> <patronymic surname>
```

Or another possible order of such set of names in GrampsWeb:
```
<surname> <patronymic surname> <given name>
```

All these order existing in GrampsWeb are looks very strange and unusual in my culture.

Lets look at some examples.

List of persons (`https://<domain>/people`):
![image](https://github.com/gramps-project/gramps-web/assets/28873347/dcf9e50c-c6ab-48c9-967e-c25206c1c289)

Info about concrete person (`https://<domain>/person/<person_id>`):
![image](https://github.com/gramps-project/gramps-web/assets/28873347/ed89331c-49e5-4e3e-8b45-c2f42d6fabb0)

Ancestor tree and other types of trees (`https://<domain>/tree`):
![image](https://github.com/gramps-project/gramps-web/assets/28873347/93b8f5f2-5e3e-4339-8408-005c026354a5)

Maybe some other views exists with same order of names unusual for my culture.

Desktop Gramps application has elegant way to solve this problem. It has setting for for custom name display format:
https://gramps-project.org/wiki/index.php/Gramps_5.2_Wiki_Manual_-_Settings#Display_Name_Editor

I can set it to desired value and after it order of names looks natively to my culture in all views I have found in desktop app. For example some screenshots from Desktop Gramps 5.2.2 after setting of custom name display format (sorry for screenshots from desktop app - they are in translation to my language, I can't find howto change translation to English in desktop app).

List of persons:
![image](https://github.com/gramps-project/gramps-web/assets/28873347/b8338385-49e7-4ba5-aa7c-410fe4e9d792)

Info about concrete person:
![image](https://github.com/gramps-project/gramps-web/assets/28873347/9eaedabe-51d4-4fef-b253-d7d2b5b09daf)

Tree:
![image](https://github.com/gramps-project/gramps-web/assets/28873347/8ab26f73-87e4-44b9-a319-bbae0f5ff234)

Please, implement ability to change default order of names in all views in GrampsWeb according to setting of custom name display format in GrampsWeb, just like works in desktop Gramps.

P.s.

Version of GrampsWeb where I created screenshots:
```
Gramps 5.2.2
[...]

PR #638, opened 2025-03-28
Add option to change name display format in charts

### LINKED ISSUE #411: Option to change display format for name in graph views
Can we have option to choose the display format for name. Currently it's defaulted to _Surname, FirstName_. 
Would be nice if there is option to choose _FirstName Surname_ as well.

Sebolains:
Fixes #411 

This PR does not attempt to implement a comprehensive, system wide name rending system as proposed by #444. This is only intended to address Charts.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -218,5 +218,8 @@
   "Log in to Gramps Web": "Log in to Gramps Web",
   "Invalid E-mail address": "Invalid E-mail address",
   "full-text": "full-text",
-  "semantic": "semantic"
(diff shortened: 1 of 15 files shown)


DavidMStraub:
Thanks a lot! I will review it tomorrow.

DavidMStraub:
Thanks for that! It works really well and is a pragmatic approach as long as we don't have anything more comprehensive. I suspect users might complain that the setting is not respected when switching the graph, but we can add this later.

PR #654, opened 2025-05-07
Toggle person outlines on photos

### LINKED ISSUE #605: option to hide person overlay on photos.
**Is your feature request related to a problem? Please describe.**
When having large group photos with many individuals, it would be nice to have the option to show/hide the person overlay -  maybe with the option to show a hover information when moving the mouse over a (hidden) person area.

I also miss the option to zoom in photos.

**Describe the solution you'd like**
a switch or similar to hide the person overlay on photos.

jdelaney72:
I added a toggle button that shows/hides (opacity: 0) the person outline rectangles in the lightbox view. When hidden, the person outline rectangles will only show when you hover over them with your mouse.

I tried a switch (mwc-switch instead of mwc-icon-button) first but it took up too much of the limited real estate in the bottom right corner.

This should fix #605.

Feedback or suggestions welcome - this is my first PR for this project.

localhost_5555_person_I0880 (1)

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -221,5 +221,6 @@
   "semantic": "semantic",
   "Name Display Format": "Name Display Format",
   "Surname First": "Surname First",
-  "Given Name First": "Given Name First"
+  "Given Name First": "Given Name First",
+  "Toggle person outlines": "Toggle person outlines"
 }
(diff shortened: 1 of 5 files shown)


DavidMStraub:
This is great! I like it almost so much as to make hiding them the default, but it is not mobile friendly as there is no hover on mobile, so I think the way you did it is the best trade-off.

Only thing to fix, please put the new string at the end of `en.json`, as it's a frontend-specific string (`strings.py` is for loading strings that exist in the Gramp backend).

jdelaney72:
Fixed!

I was looking for the translations but I guess I missed the `lang` folder.

Issue #470, opened 2024-07-17
The "new attribute" modal for editing a person is too small


gustavi:
**Description**

The "new attribute" modal for editing a person is too small and hard to use. The select do not display more than 2 full options.

![image](https://github.com/user-attachments/assets/0b7d0184-cb64-47c4-bbb6-6ff88a11fbc3)

**How to reproduce**

- Go to a person profile (`/person/I[N]`)
- Edit (bottom right)
- Attributes tab
- "+" button

**Proposed fixes**

1. Making the modal taller
2. Allow the select to go beyond the modal

**Additional information**

- Firefox 115
- 1920x1080
- Gramps Web Frontend 24.7.0

DavidMStraub:
Right. But no need to enlarge the modal, just need to set `fixedMenuPosition` on the `mwc-select`.

DavidMStraub:
Turns out this can't be fixed easily with MWC, see https://github.com/material-components/material-web/issues/832.

The solution is to migrate to `md-filled-select` which has a `menuPositioning` attribute, see https://material-web.dev/components/select/. But is more work.

DavidMStraub:
Also true for place types:

https://github.com/gramps-project/gramps-web/issues/481#issuecomment-2277871730

m-breitbach:
I observed that, if the dropdown menu opens downwards and it's truncated, the whole modal can be scrolled down at least (this makes it even usable, even if not super conveniently). In cases where it openes upwards and is truncated (like in place types), no such scrolling becomes availabe.

DavidMStraub:
Not in progress, would be great if you could tackle it. The solution is to migrate from `mwc-dialog` and `mwc-select` to `md-dialog` and `md-filled-select`, there is no point in investing work into workarounds for the `mwc-` elements as we need to get rid of them anyway.

Step 1: replace `mwc-dialog` in `GrampsjsObjectForm.js` by `md-dialog`. `md-dialog` is already used in some other places in the code base, you can look at those.

Step 2: replace `mwc-select` in `GrampsjsFormSelectType.js` by `md-filled-select`. Also `md-filled-select` is already used in two places.

Note: I am not actually sure if the migration fixes the problem by itself, but if it doesn't, we should find a fix/workaround for the new elements rather than the unmaintained ones.

m-breitbach:
Already experimented with workarounds for `mwc-`, but I failed miserably due to uncontrollable side effects. So your approach definitely makes sense.

[...]

PR #662, opened 2025-05-15
Replace some MWC with MD components

m-breitbach:
Addressing #470.

Please check my comments.

### DIFF
--- a/src/SharedStyles.js
+++ b/src/SharedStyles.js
@@ -17,6 +17,14 @@ export const sharedStyles = css`
     --mdc-typography-font-family: var(--grampsjs-body-font-family);
     --color-boy: #64b5f6;
     --color-girl: #ef9a9a;
+    --md-sys-typescale-headline-small-font: var(--grampsjs-heading-font-family);
+    --md-sys-typescale-headline-medium-font: var(
+      --grampsjs-heading-font-family
+    );
+    --md-sys-typescale-headline-large-font: var(--grampsjs-heading-font-family);
+    --md-sys-typescale-body-small-font: var(--grampsjs-body-font-family);
+    --md-sys-typescale-body-medium-font: var(--grampsjs-body-font-family);
+    --md-sys-typescale-body-large-font: var(--grampsjs-body-font-family);
     --md-sys-color-primary: rgb(109 76 65);
     --md-sys-color-surface-tint: rgb(109 76 65);
     --md-sys-color-on-primary: rgb(255 255 255);
@@ -264,6 +272,12 @@ export const sharedStyles = css`
     color: var(--mdc-theme-secondary);
     --mdc-theme-primary: var(--mdc-theme-secondary);
     --mdc-theme-on-primary: var(--mdc-theme-on-secondary);
+    --md-sys-color-primary: var(--md-sys-color-secondary);
+    --md-sys-color-on-primary: var(--md-sys-color-on-secondary);
+    --md-sys-color-primary-container: var(--md-sys-color-secondary-container);
+    --md-sys-color-on-primary-container: var(
+      --md-sys-color-on-secondary-container
+    );
   }
 
   mwc-icon-button.large {
--- a/src/SharedStyles.js
+++ b/src/SharedStyles.js
@@ -73,7 +73,7 @@ export const sharedStyles = css`
     --md-sys-color-surface-container-low: rgb(255 241 237);
     --md-sys-color-surface-container: #fff;
     --md-sys-color-surface-container-high: #fff;
-    --md-sys-color-surface-container-highest: #fff;
+    --md-sys-color-surface-container-highest: whitesmoke;
     --md-primary-tab-label-text-weight: 425;
     --md-primary-tab-label-text-size: 16px;
     --md-primary-tab-active-indicator-height: 3px;
--- a/src/components/GrampsjsFormSelectType.js
+++ b/src/components/GrampsjsFormSelectType.js
@@ -3,8 +3,8 @@ Element for selecting a Gramps type
 */
 
 import {css, html, LitElement} from 'lit'
-import '@material/mwc-select'
-import '@material/mwc-list/mwc-list-item'
+import '@material/web/select/filled-select.js'
+import '@material/web/select/select-option.js'
 
 import {classMap} from 'lit/directives/class-map.js'
 import {sharedStyles} from '../SharedStyles.js'
@@ -76,12 +76,12 @@ class GrampsjsFormSelectType extends GrampsjsAppStateMixin(LitElement) {
         ? ''
         : html`<h4 class="label">${this.heading || this._('Type')}</h4>`}
       <p style="display: flex">
-        <mwc-select
+        <md-filled-select
           style="width:100%"
           class="${classMap({hide: this._hasCustomType})}"
           ?required=${this.required && !this._hasCustomType && !this.nocustom}
           ?disabled=${this.loadingTypes || this._hasCustomType || this.disabled}
-          validationMessage="${this._('This field is mandatory')}"
+          error-text="${this._('This field is mandatory')}"
           @change="${this.handleChange}"
           label="${this.loadingTypes ? this._('Loading items...') : this.label}"
           id="select-type"
@@ -89,21 +89,24 @@ class GrampsjsFormSelectType extends GrampsjsAppStateMixin(LitElement) {
           ${types.includes(this.defaultTypeName) ||
           types.includes(this.initialValue)
             ? ''
-            : html`<mwc-list-item value="" selected></mwc-list-item>`}
+            : html`<md-select-option value="" selected
+                ><div slot="headline"></div
+              ></md-select-option>`}
           ${this.loadingTypes
             ? ''
             : types.map(
                 (obj, i) => html`
-                  <mwc-list-item
+                  <md-select-option
                     value="${this.valueNonLocal ? types[i] : typesLocale[i]}"
                     ?selected="${(this.initialValue &&
                       obj === this.initialValue) ||
                     (!this.initialValue && obj === this.defaultTypeName)}"
-                    >${this._(obj)}</mwc-list-item
                   >
+                    <div slot="headline">${this._(obj)}</div>
+                  </md-select-option>
                 `
               )}
-        </mwc-select>
+        </md-filled-select>
         ${this.nocustom || !this._hasCustomType
           ? ''
           : html`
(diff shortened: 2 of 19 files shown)


src/components/GrampsjsFormSelectType.js
```diff
@@ -3,8 +3,8 @@ Element for selecting a Gramps type
 */
 
 import {css, html, LitElement} from 'lit'
-import '@material/mwc-select'
-import '@material/mwc-list/mwc-list-item'
+import '@material/web/select/filled-select.js'
```
m-breitbach, line 6: Does it even need these imports? Seems to work without in local test setup, don't know why.

src/components/GrampsjsFormSelectType.js
```diff
@@ -76,34 +76,37 @@ class GrampsjsFormSelectType extends GrampsjsAppStateMixin(LitElement) {
...
         : html`<h4 class="label">${this.heading || this._('Type')}</h4>`}
       <p style="display: flex">
-        <mwc-select
+        <md-filled-select
           style="width:100%"
           class="${classMap({hide: this._hasCustomType})}"
           ?required=${this.required && !this._hasCustomType && !this.nocustom}
           ?disabled=${this.loadingTypes || this._hasCustomType || this.disabled}
-          validationMessage="${this._('This field is mandatory')}"
+          error-text="${this._('This field is mandatory')}"
```
m-breitbach: With the new component, the error text only shows in case of empty custom text, and doesn't if only the empty list element is selected. It maybe needs custom logic to get the old behavior.

src/components/GrampsjsObjectForm.js
```diff
@@ -36,12 +33,10 @@ export class GrampsjsObjectForm extends GrampsjsAppStateMixin(LitElement) {
           color: rgba(0, 0, 0, 0.5);
         }
 
-        mwc-dialog {
-          --mdc-dialog-max-width: 100vw;
-          --mdc-dialog-min-width: 50vw;
-          --mdc-dialog-content-ink-color: var(
```
m-breitbach, line 42: Don't know if still required. Also, better manage all `md` attributes globally?

src/components/GrampsjsObjectForm.js
```diff
@@ -36,12 +33,10 @@ export class GrampsjsObjectForm extends GrampsjsAppStateMixin(LitElement) {
...
-            --mdc-theme-text-primary-on-background
-          );
+        md-dialog {
+          max-width: 100vw;
+          min-width: 50vw;
+          max-height: 80vh;
```
m-breitbach, line 39: My attempt to mimic the old height-sizing behavior.

src/components/GrampsjsObjectForm.js
```diff
@@ -130,37 +125,24 @@ export class GrampsjsObjectForm extends GrampsjsAppStateMixin(LitElement) {
...
   render() {
     return html`
-      <mwc-dialog
-        scrimClickAction=""
-        escapeKeyAction=""
-        heading="${this.dialogTitle}"
-        open
-      >
-        <div @formdata:changed="${this._handleFormData}">
+      <md-dialog @cancel="${this._handleDialogCancel}" open>
```
m-breitbach, line 128: `@cancel` captures the ESC button. (Ignoring it I consider unfriendly UX.)

src/SharedStyles.js
```diff
@@ -17,6 +17,14 @@ export const sharedStyles = css`
...
     --color-girl: #ef9a9a;
+    --md-sys-typescale-headline-small-font: var(--grampsjs-heading-font-family);
+    --md-sys-typescale-headline-medium-font: var(
+      --grampsjs-heading-font-family
+    );
+    --md-sys-typescale-headline-large-font: var(--grampsjs-heading-font-family);
+    --md-sys-typescale-body-small-font: var(--grampsjs-body-font-family);
+    --md-sys-typescale-body-medium-font: var(--grampsjs-body-font-family);
+    --md-sys-typescale-body-large-font: var(--grampsjs-body-font-family);
     --md-sys-color-primary: rgb(109 76 65);
```
m-breitbach, line 28: I noticed that the tones (brown and blue) differ a bit from the `mdc-` values. Is this desired?

DavidMStraub: I don't remember to be honest, but I think it's ok.

src/components/GrampsjsFormSelectType.js
DavidMStraub: I am also not sure why it works :sweat_smile: I think it is because it is imported by a parent component, but I think it's good practice to import all the requirements in each component - in principle they could be used standalone. In the final build, I think it doesn't matter anyway.

src/components/GrampsjsFormSelectType.js
DavidMStraub: The important thing is just that the final "add" button is still grayed out if the empty element is selected, which I think is still the case, right?

src/components/GrampsjsObjectForm.js
DavidMStraub: For the color, definitely yes, better to manage globally. For the width/height, I think it's ok to do it here as different dialogs can have different preferred sizes.

src/components/GrampsjsFormSelectType.js
m-breitbach: Do you mean the "Save" button? Neither in my implementation nor previously was it grayed out in case of empty costom string or empty selection. 

(Below screenshot is from an instance running the latest front end version.)

In any case, we should probably do all that properly from now on (the error message and the gray-out). Do we really need the empty list entry in case of required fields? We could just use the first entry as a default.

image

dstraub-pf: Ah, this is the object form version sorry, the button *is* grayed out in the "new event" stand-alone view. I think it would make sense if we implement the same behaviour here.

> Do we really need the empty list entry in case of required fields? We could just use the first entry as a default.

Yes, good idea!

DavidMStraub: Ah, this is the object form version sorry, the button is grayed out in the "new event" stand-alone view. I think it would make sense if we implement the same behaviour here.

> Do we really need the empty list entry in case of required fields? We could just use the first entry as a default.

Yes, good idea!

m-breitbach: Great, I will try to implement it like that in the next few days.

m-breitbach: I checked the stand-alone version, and its behavior is unchanged. However, I noticed a bug there: When entering custom text mode, adding text (ungraying the "Add"-button), and then cancelling custom text mode (switching back to the "empty" list entry), the "Add"-button stays ungrayed. Anyways, this will be mitigated with the removal of the empty entry for required fields.

DavidMStraub: By the way, Gramps desktop has some logic to select the "most likely" type as default. If no birth event exists, it uses birth, if it does, death, etc. Not sure we want to go there, but just in case you feel like it :smile:

DavidMStraub:
Please ping me when you think it's ready to merge.

m-breitbach:
I figured that a consistent behavior in terms of validity and error messages is not possible with `md-` elements out of the box, but requires custom logic. The select vs. custom functionality makes it pretty complex with many possible states and actions. I ended up in a broader rewrite of the logic of that class. I am almost there (behavior is already more consistent than before) but still need to fix some issues.

src/components/GrampsjsFormSelectType.js
m-breitbach: I would postpone implementing the button grey-out behavior inside the dialogs to another PR, as it wasn't done correctly already before. At least the buttons in the new-event-case schould now work robustly.

DavidMStraub:
I'm testing your branch locally, looks great overall, smaller issues:

- Tooltips are hidden behind the modal, looks like a z-index problem

grafik

- When adding a new event, I still see the empty list element as an option, is that intentional? I though you wanted to select the first non-empty one by default.
- The save button is not grayed out for invalid (empty) input, let me know if this is out of scope of your PR, then I'll open an issue so we don't forget.

m-breitbach:
- I looked into it for a while now and didn't find a way to fix. I played around with z-values in various places, but it didn't help. Thats where my front end knowledge ends. AI suggests that it might help to either remove the stacking context of the dialog (?) or render the tippy as a child of the dialog rather than `<body>`... Is there a chance you could give that a try? I guess you know the app much better than I do. (Sorry!)
- I tried to make the blank entry depend on the initial value. However, the latter is dynamically bound to `.data`, so it doesn't work the naive way. I tried to solve it using Lit's `firstUpdated()` and `updated()` but it made the logic complicated and uncontrollable. (I've spend many hours...) Finally, I decided to leave it this way, which is at least perfectly predictable. If you still find a good way to do it, I am happy do learn!
- Yes, I would see the grey-out of the button inside the pop-up dialog out of scope, leaving the behavior as it was. At least in the "New Event" page, the grey-out behavior now works a bit more consistently (there was a small bug, as I described in the other conversation).

DavidMStraub:
Thanks for the explanations!

I also tried to un-hide the tooltip but failed. This would be a bit of a showstopper as the "+" icon button is not really understandable without tooltip. But perhaps that's anyway a sign of poor UI: tooltips are not mobile friendly.

So what about just adding a text button below? This would circumvent the issue and be clearer.

grafik


```javascript
  render() {
    return html`
      ${this.noheading ? '' : this.#renderHeading()}
      <p>
        ${this.loadingTypes ? this.#renderLoading() : this.#renderInputs()}
        ${this.nocustom ? '' : this.#renderCustomSwitch()}
      </p>
    `
  }

...
  #renderCustomSwitch() {
    return html`
      <md-text-button
        style="margin-top: 8px;"
        id="button-switch-type"
        @click="${this.#toggleCustomType}"
        ?disabled="${this.disabled || this.loadingTypes}"
        >${this._hasCustomType
          ? this._('Switch to default type')
          : this._('Add custom type')}</md-text-button
      >
    `
  }
```

m-breitbach:
I like your idea, added it.

m-breitbach:
By the way, I notice that the MWC components had a background color, which the MD didn't have in the current setup. See my latest commit, which mimics the old behavior. However, unsure if we/you want that. I have no opinition. (Let me know, if I should drop that commit.)

DavidMStraub:
Hmm, I don't have a strong opinion either. Let's use it like that.

Let's ship this! :rocket:

m-breitbach:
Awesome :) I think after the full transition to MD, one could unify the whole color scheme.

PR #713, opened 2025-08-06
Allow alternative place name to be edited, added and removed

cajturner:
https://github.com/gramps-project/gramps-web/issues/27

Allow place alt names to be added, edited and removed.

image

### DIFF
--- a/src/components/GrampsjsFormEditPlaceName.js
+++ b/src/components/GrampsjsFormEditPlaceName.js
@@ -5,6 +5,7 @@ Form for adding a new event reference
 import {html} from 'lit'
 
 import './GrampsjsFormString.js'
+import './GrampsjsFormSelectDate.js'
 import {GrampsjsObjectForm} from './GrampsjsObjectForm.js'
 
 class GrampsjsFormEditPlaceName extends GrampsjsObjectForm {
@@ -18,7 +19,25 @@ class GrampsjsFormEditPlaceName extends GrampsjsObjectForm {
         value="${this.data.value || ''}"
       >
       </grampsjs-form-string>
+      <grampsjs-form-select-date
+        @formdata:changed="${this._handleFormData}"
+        id="place-name-date"
+        label="${this._('Date')}"
+        .data="${this.data.date || {dateval: ''}}"
+        .appState="${this.appState}"
+      >
+      </grampsjs-form-select-date>
     `
+  }
+
+  _handleFormData(e) {
+    const originalTarget = e.composedPath()[0]
+    if (originalTarget.id === 'place-name-value') {
+      this.data = {...this.data, value: e.detail.data}
+    } else if (originalTarget.id === 'place-name-date') {
+      this.data = {...this.data, date: e.detail.data}
+    }
+    super._handleFormData(e)
   }
 }
 
--- a/src/components/GrampsjsFormEditPlaceName.js
+++ b/src/components/GrampsjsFormEditPlaceName.js
@@ -9,6 +9,18 @@ import './GrampsjsFormSelectDate.js'
 import {GrampsjsObjectForm} from './GrampsjsObjectForm.js'
 
 class GrampsjsFormEditPlaceName extends GrampsjsObjectForm {
+  static get properties() {
+    return {
+      ...super.properties,
+      showDateInput: {type: Boolean},
+    }
+  }
+
+  constructor() {
+    super()
+    this.showDateInput = false
+  }
+
   renderForm() {
     return html`
       <grampsjs-form-string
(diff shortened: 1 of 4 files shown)


DavidMStraub:
Nice!

Btw I noticed we are not even using the `profile.alternate_place_names` yet, that you added to the backend, right?

DavidMStraub:
Needs rebase.

cajturner:
@DavidMStraub I have rebased and fixed a few issues. Let me know what you think

src/components/GrampsjsFormEditPlaceName.js
```diff
@@ -5,9 +5,22 @@ Form for adding a new event reference
 import {html} from 'lit'
 
 import './GrampsjsFormString.js'
+import './GrampsjsFormSelectDate.js'
 import {GrampsjsObjectForm} from './GrampsjsObjectForm.js'
 
 class GrampsjsFormEditPlaceName extends GrampsjsObjectForm {
+  static get properties() {
+    return {
+      ...super.properties,
```
DavidMStraub: There's need for that, see [Lit docs](https://lit.dev/docs/components/properties/): 

> You don't need to redeclare properties unless you want to change options.

src/components/GrampsjsObject.js
```diff
@@ -671,7 +669,7 @@ export class GrampsjsObject extends GrampsjsAppStateMixin(LitElement) {
               >
                 <grampsjs-map-overlay
                   url="${getMediaUrl(this.data.handle)}"
-                  bounds="${mapBounds[0].value}"
+                  .bounds="${mapBounds[0].value}"
```
DavidMStraub: I think this doesn't work - I know it throws a linter error, but for some reason this needs to be passed as string rather than array (i.e. without dot). Or did you observe something else?

src/components/GrampsjsFormEditPlaceName.js
```diff
@@ -18,8 +31,30 @@ class GrampsjsFormEditPlaceName extends GrampsjsObjectForm {
         value="${this.data.value || ''}"
       >
       </grampsjs-form-string>
+      ${this.showDateInput
+        ? html`
+            <grampsjs-form-select-date
+              @formdata:changed="${this._handleFormData}"
+              id="place-name-date"
+              label="${this._('Date')}"
+              .data="${this.data.date || {dateval: ''}}"
```
DavidMStraub: This is not a valid date because `dateval` needs to be an array, or am I missing something?

Perhaps you want `emptyDate` from `util.js`?

Also, is `||` the appropriate operator here? If `this.data.date` is a date, I guess you either want `??` if you only need to checked whether it's undefined, or you need `isDateEmpty` if for some reason you need to check whether it's empty (I didn't check in detail).

src/components/GrampsjsObject.js
```diff
@@ -591,34 +591,32 @@ export class GrampsjsObject extends GrampsjsAppStateMixin(LitElement) {
           ></grampsjs-names>
         `
       case 'placeNames':
-        return html` ${this.data.alt_names?.length > 0
+        return html` ${this.data.alt_names?.length > 0 || this.edit
           ? html` <grampsjs-place-names
               .appState="${this.appState}"
               .strings="${this.strings}"
-              .data="${this.data.alt_names}"
-              .profile="${this.data.profile.alternate_place_names ?? []}"
```
DavidMStraub, line 599: I added this so we actually use the formatted date from the backend

DavidMStraub:
It looks good, I added a few comments - in particular in rebasing it seems you accidentally removed the code that uses the formatted date from the profile.

src/components/GrampsjsObject.js
cajturner: Ah we need to parse this before passing. Resolves the linting issue and still works 😅 
image

src/components/GrampsjsFormEditPlaceName.js
```diff
@@ -18,8 +31,34 @@ class GrampsjsFormEditPlaceName extends GrampsjsObjectForm {
...
+              .appState="${this.appState}"
+            >
+            </grampsjs-form-select-date>
+          `
+        : ''}
     `
   }
+
+  get isValid() {
+    return this.data.value !== ''
```
cajturner: Lets ensure the name is set. Date can be empty

src/components/GrampsjsObject.js
```diff
@@ -671,7 +669,7 @@ export class GrampsjsObject extends GrampsjsAppStateMixin(LitElement) {
               >
                 <grampsjs-map-overlay
                   url="${getMediaUrl(this.data.handle)}"
-                  bounds="${mapBounds[0].value}"
+                  .bounds="${JSON.parse(mapBounds[0].value)}"
```
cajturner, line 673: Resolve linting error

src/components/GrampsjsPlaceNames.js
```diff
@@ -3,44 +3,117 @@ import {html} from 'lit'
...
 import {GrampsjsEditableTable} from './GrampsjsEditableTable.js'
+import './GrampsjsFormEditPlaceName.js'
 
 export class GrampsjsPlaceNames extends GrampsjsEditableTable {
-  static get properties() {
-    return {
-      profile: {type: Array},
-    }
-  }
-
```
cajturner, line 16: Lets pass the profile array into the data instead

cajturner:
> It looks good, I added a few comments - in particular in rebasing it seems you accidentally removed the code that uses the formatted date from the profile.

Thanks @DavidMStraub, great comments. With fresh eyes I can see a few other improvements. Should be ready for a re-review now

It's looking much better now
image
image

src/components/GrampsjsObject.js
DavidMStraub: Haha, excellent point! :man_facepalming:

DavidMStraub:
Perfect! :partying_face:

PR #714, opened 2025-08-06
Ensure selects overflow when in dialog

cajturner:
I noticed select options don't overflow the dialog. I thought it would be a nice visual improvement if they did. 

I think I have updated all the selects that are in dialogs, but not 100% sure

### Before
Screenshot 2025-08-06 at 22 32 04

### After 
Screenshot 2025-08-06 at 22 31 37


DavidMStraub:
Perfect, thanks!

I still don't know what the equivalent of this with the new `md-select` is :thinking:

Issue #145, opened 2022-12-07
Darkmode for Web Interface


onlynow2:
It would be nice, having the option for toggeling light / dark mode.
Possible location would be in the settings menu or next to the search / profile icon top right.

DavidMStraub:
I agree this would be something that some users might find useful. Personally, I will not work on it since it's a huge amount of work for a feature I will not use. If someone wants to tackle this, please self-assign and I can provide guidance if needed.

zeitchef:
I might be interested in looking into this. @DavidMStraub Are there any guidelines regarding dark mode design tokens, or would this still need to be sorted?

DavidMStraub:
Hi, that would be great! There are no guidelines yet. However, over the next month, I expect we will slowly start migrating the components from the legacy material components v0.25 to the new material web v1.0, which has buil-in [support for themes](https://material-web.dev/theming/material-theming/) as far as I can see. Perhaps we can already start using those conventions? Does that help?

At the moment, the main colors are set here using the CSS variables of the old material components:
https://github.com/gramps-project/Gramps.js/blob/main/src/SharedStyles.js#L10-L15

However, there are still many components that have hard-coded colors (e.g. shades of gray), that will have to be adapted.

Some suggestions for implementing the dark mode switch:

- We can store the user preference in local storage using the existing functions `updateSettings` and `getSettings`
- I suggest to put the theme selection on the user settings page where also the language selector is located

zeitchef:
Hi @DavidMStraub - I've got some time this week and would like to dig into this a bit. I've hit my first glitch though, after setting up a dev environment, I can't seem to register a test user - every combination I've tried just returns a `409` error. Is there a dedicated test user I can use to login with, or a pattern I can use to register one? Thanks.

DavidMStraub:
(Confusingly, in the new Material Web, components and CSS variables start with `md-`, while in the old version components started with `mwc-` and CSS variables with `mdc-`.)

tismofied:
Has there been any progress made on this?

DavidMStraub:
No.

joelabc:
Would be a nice to have NGL

DavidMStraub:
Implemented in #755!

PR #755, opened 2025-09-17
Dark mode (#145)

mschuepbach:
I made an attempt to add a dark mode (#145). 

The functionality seems to be working; it uses the system theme by default, changes the theme when the system theme changes and a theme can be explicitly set in the user settings.

The main remaining issue are the older material components. Styling them is a bit tedious. Would it make more sense to replace them with the newer material components first? I could try that in a separate PR, but I’m not sure which approach would be easier.

Another issue is a white flash when the page loads.

Also, I am not a designer, I just generated the colors with https://material-foundation.github.io/material-theme-builder/.


DavidMStraub:
Brilliant, thank you! Let's do this.

Can you please make one change (or convince me otherwise :wink:): please let's only put CSS variables for colors into `index.html` and leave the rest (especially typography) in `sharedStyles`, ok?

> The main remaining issue are the older material components. Styling them is a bit tedious. Would it make more sense to replace them with the newer material components first? I could try that in a separate PR, but I’m not sure which approach would be easier.

I also thought let's first switch them all over, but I realized it's a lot of work for buttons and lists, so I think this shouldn't block this feature. Perhaps feeding the color token docs of both libraries will make it easier to translate the CSS variables into each other to automate this?

> Another issue is a white flash when the page loads.

You mean until it has detected the system's dark mode settings? Does it also happen when dark mode is selected explicitly in the user settings (so it's persisted in localStorage)?

> Also, I am not a designer, I just generated the colors with https://material-foundation.github.io/material-theme-builder/.

Neither am I, I think it's a fine approach, plus tweaking manually what looks weird.

mschuepbach:
> Can you please make one change (or convince me otherwise 😉): please let's only put CSS variables for colors into `index.html` and leave the rest (especially typography) in `sharedStyles`, ok?

Ah yeah, sure. This was just to get it working quickly. I could also move the styles into a separate css file so that `index.html` doesn't get cluttered.

> I also thought let's first switch them all over, but I realized it's a lot of work for buttons and lists, so I think this shouldn't block this feature. Perhaps feeding the color token docs of both libraries will make it easier to translate the CSS variables into each other to automate this?

Okay. What do you mean with "color token docs"? My idea was to assign a `--md-sys-color-*` variable to every relevant `--mdc-*` variable so that only one variable needs to be updated if you want to change a color. How would you automate these assignments?

> You mean until it has detected the system's dark mode settings? 

Yes. Of course this is not a problem if the user uses light mode. This is known as "[Flash of inAccurate coloR Theme (FART)](https://css-tricks.com/flash-of-inaccurate-color-theme-fart/)"

> Does it also happen when dark mode is selected explicitly in the user settings (so it's persisted in localStorage)?

Yes. I think the detection of the theme just needs to happen earlier.

Should the light theme keep all the exact colors as we currently have? Because my idea was to only use the `--md-sys-color-*` variables. But it didn't look very good when I replaced some of the hardcoded values with those variables.

DavidMStraub:
> What do you mean with "color token docs"? My idea was to assign a --md-sys-color-* variable to every relevant --mdc-* variable so that only one variable needs to be updated if you want to change a color. How would you automate these assignments?

Yes, exactly. Not sure if it's possible to automate anything, I was thinking of docs like [this one](https://material-web.dev/components/button/#outlined-button-tokens), perhaps an LLM can help to come up with a not terrible starting point.

> Yes. Of course this is not a problem if the user uses light mode. This is known as "[Flash of inAccurate coloR Theme (FART)](https://css-tricks.com/flash-of-inaccurate-color-theme-fart/)"

Thanks, I'll add that to my vocabulary :laughing: 

> Should the light theme keep all the exact colors as we currently have? Because my idea was to only use the --md-sys-color-* variables. But it didn't look very good when I replaced some of the hardcoded values with those variables.

Not necessarily exact, but no drastic visual changes for the moment I'd say.

Note that most of these colors here https://github.com/gramps-project/gramps-web/blob/main/src/SharedStyles.js#L32-L77 (namely the ones with `rgb` without commas) were also generated by some color scheme generator when I switched to `md` components for the first time, so many of those might not be actually needed and/or pretty.  Basically anytime I migrate one of the components, I change the colors in the stylesheet that look bad.

DavidMStraub:
Created the following with the help of copilot. Those are only the MWC theme colors, not the component specific ones, so not sure how far it will get us. (And I haven't tested it.)

```css
/* Bridge MWC theme variables to Material Design CSS color variables */
:root {
  --mdc-theme-primary: var(--md-sys-color-primary);
  --mdc-theme-secondary: var(--md-sys-color-secondary);
  --mdc-theme-surface: var(--md-sys-color-surface);
  --mdc-theme-background: var(--md-sys-color-background);
  --mdc-theme-error: var(--md-sys-color-error);
  --mdc-theme-on-primary: var(--md-sys-color-on-primary);
  --mdc-theme-on-secondary: var(--md-sys-color-on-secondary);
  --mdc-theme-on-surface: var(--md-sys-color-on-surface);
  --mdc-theme-text-primary-on-background: var(--md-sys-color-on-background);
  --mdc-theme-text-secondary-on-background: var(--md-sys-color-on-surface-variant);
  --mdc-theme-text-hint-on-background: var(--md-sys-color-on-surface-variant);
  --mdc-theme-text-disabled-on-background: var(--md-sys-color-on-surface-variant);
  --mdc-theme-text-icon-on-background: var(--md-sys-color-on-surface-variant);
}
```

mschuepbach:
Is there an easy way to view the DNA matches and LLM chat to check if my changes look good there?

DavidMStraub:
Good question, I need to check what the best way is.

Is that all that's still missing?

mschuepbach:
Not quite, but there is also not that much left.

**1. Dark mode primary color**
I currently have the same primary color for `--mdc-theme-primary` in both themes, but the contrast is bad for dark mode. I could just do `--mdc-theme-primary: var(--md-sys-color-primary)` but I'm not sure if having this light color on the top app bar looks good.

image


I tried this but that didn't work:
```css
:root[data-theme='dark'] mwc-top-app-bar {
  --mdc-theme-primary: var(--md-sys-color-primary-container);
}
```

I guess i could just add a separate variable. That would look like this:
image


**2. Fan chart colors**
I need to fix some colors in the fan chart.
image

**3. User-defined tag colors**
Tag colors that users created in light mode might have a low contrast in dark mode and vice-versa.
image
image


**4. Map theme**
This is optional, but i saw that there is a theme [plugin](https://github.com/lhapaipai/maplibre-theme) for maplibre that supports dark mode. 

**5. Testing**
There might be more contrast issues I haven't discovered yet.

DavidMStraub:
Great progress!

As for the app bar, I would definitely make it just dark grey as in the examples [here](https://m2.material.io/design/color/dark-theme.html) or [here](https://m3.material.io/blog/android-dark-theme-tutorial).

DavidMStraub:
Instructions for using chat in the dev setup:

1. Add `OPENAI_API_KEY: your-key` to `.devcontainer/docker-compose.yml`
2. Add `GRAMPSWEB_LLM_MODEL: gpt-4o-mini`
3. Rebuild the dev container
4. Go to user settings in Gramps Web and select the user groups that can use chat
5. Logout and login again (note: that this is needed is silly. I will open an issue)

mschuepbach:
Thank you. The chat colors look good.

DavidMStraub:
Looks really good overall!

The map style also works by just replacing `liberty` by `dark` in `GrampsjsMap.js` – the only non-trivial part is handling the style change event I guess.

DavidMStraub:
I find this brown a bit prettier, what do you think?

image

```css
  --md-sys-color-primary: #8d7168;
  --md-sys-color-on-primary: rgba(0, 0, 0, 0.8);
```

mschuepbach:
Okay, sure. Looks good to me.

Another small thing I want to change is the elevation for dialogs. The difference between dialogs and the background is barely visible.

DavidMStraub:
Btw, to make the theme selector translatable, we also need to add the new strings "System", "Light", and "Dark" to `en.json`.

DavidMStraub:
Let me know when you think it's ready to :ship:!

DavidMStraub:
Awesome work! :partying_face:

Issue #748, opened 2025-09-10
Improve Temporal Place Hierarchy Management


Ziesie1:
Gramps Web currently allows assigning a place to a higher-level location for a specific time period (e.g. Celle belonging to Landkreis Celle between certain years). However, once this hierarchy is created, it cannot be edited, and the time period is not visible anywhere in the UI.

This makes it difficult to manage historical accuracy, especially when administrative boundaries change over time.

<img width="2090" height="1646" alt="Image" src="https://github.com/user-attachments/assets/02faeee4-44f0-46c0-ab67-ee3fb405ff8d" />

<img width="579" height="775" alt="Image" src="https://github.com/user-attachments/assets/7f794b16-618a-42a2-b14c-f0d86badaa2e" />

**Suggested improvements:**
- Allow users to edit existing place hierarchy relationships, including the time period.
- Display the assigned time period in the place hierarchy view.

Thanks for your great work on Gramps Web!

PR #757, opened 2025-09-19
Add editing of place references (#748)

Ziesie1:
This PR implements part of the changes requested in issue #748.

- Adds support for editing existing place hierarchy relationships

⚠️ Note: This adresses only part of #748. The feature to "display the assigned time period in the hierarchy view"  is not yet implemented and remains open. Implementing this will likely also require changes to the Gramps Web API.

Additional changes:
- Renamed the text key "Share an existing place" to "Add place reference" for clarity. This text key is not yet known in webblate,  so there should be no problems.
- Added a new text key "Edit place reference". This needs to be added to Weblate (I am currently not sure how to do this).


src/components/GrampsjsPlaceRefs.js
```diff
@@ -111,6 +127,20 @@ export class GrampsjsPlaceRefs extends GrampsjsEditableTable {
...
   }
 
+  _handleEditClick(handle) {
+    const obj = this.data.find(p => p.ref === handle)
+    this.dialogContent = html`
+      <grampsjs-form-placeref
+        @object:save="${e => this._handlePlaceRefUpdate(e, obj)}"
+        @object:cancel="${this._handlePlaceRefCancel}"
+        .appState="${this.appState}"
+        .data="${obj}"
```
DavidMStraub: Don't you also have to set the `.place` property for the selected place to be displayed correctly?

DavidMStraub:
Thanks! Looks good (just 1 comment).

Regarding the translation - you just need to add `"my string": "my string",` to `en.json`.

I agree that "Share an existing place" is not clear, but "Add place reference" is only clear to somebody who knows the Gramps data model (us), which defines that a place can reference another place it is enclosed in (parent). For users, I think this is not super clear. What about "Add reference to enclosing place"?

Ziesie1:
Thanks for the feedback! I’ve added the .place property as suggested.  
For the dialog title I went with “Add enclosing place”, as it feels short and concise – does that sound good to you?

DavidMStraub:
>  For the dialog title I went with “Add enclosing place”, as it feels short and concise – does that sound good to you?

Yes probably you're right that concise is better. I suggested the longer version because we're not adding a new place, and also the date isn't a property of the place but of the reference. But maybe I'm thinking about the data model too much again. Do you think "Link enclosing place" makes sense?

Ziesie1:
Yes, I understand what you mean. I’ll go with “Link to enclosing place” instead, as it reads a bit more natural in English.  
I’d also use the same title for both adding and editing, to keep things consistent. Does that sound good to you?

DavidMStraub:
Sounds good!

Issue #797, opened 2025-10-25
[Bug] Users cannot log in with password in v25.10


DavidMStraub:
### Bug Description

This also affects the demo https://demo.grampsweb.org/

Now investigating

@AlexBocken FYI I suspect it's related to the frontend changes.

### Gramps Web Version Information

v25.10.0.1 & Web API 3.4.0

### Verification

- [x] I have searched for existing issues before creating this one

DavidMStraub:
I found the problem. Fixing...

DavidMStraub:
Fixed by #798.

PR #798, opened 2025-10-25
Fix login form

DavidMStraub:
The login form was unfortunately broken in v25.10 due to a weird regression. This should fix it.

Background: the form's native handler and our own Javscript handler were both firing, leading to an extra reload in the wrong moment.

I'm not sure why this didn't happen before or why I didn't see it in local testing... :thinking:


DavidMStraub:
Fixes #797.

PR #807, opened 2025-10-30
fix: disable registration on backend metadata

qugu:
Do not show the registration form at all if registration was disabled on the backend.


DavidMStraub:
Have you even tried this locally? This won't work. Even if you added the information to the metadata endpoint, that endpoint needs authentication, so it cannot be queried before signing in.

qugu:
This wasn't meant to be a review item, I was just experimenting. Thanks!

PR #818, opened 2025-11-04
Add color for mwc-icon-button in meta slot

Ziesie1:
This pull request fixes an issue where the Edit button for the main person was not visible in dark mode (the icon appeared black instead of white).
The problem was caused by Material Design’s default styles not adapting to our dark mode theme.

I added a style for mwc-icon-button[slot='meta'] in the shared styles, so the icon color is now controlled by our own CSS variable.
This ensures the button remains visible in dark mode and is consistent with our theme.

I’m not a CSS expert, but this solution works for our use case and overrides the Material Design style that did not support our dark mode.

Before:
image

After:
image

### DIFF
--- a/src/SharedStyles.js
+++ b/src/SharedStyles.js
@@ -372,4 +372,8 @@ export const sharedStyles = css`
       font-size: 18px;
     }
   }
+
+  mwc-icon-button[slot='meta'] {
+    color: var(--grampsjs-color-icon);
+  }
 `


DavidMStraub:
Hi @Ziesie1, I don't know why, but the icon is now invisible in light mode. Can you please fix it? Thanks!

Ziesie1:
Hi,
sorry for the issue. I actually tested it in light and dark mode (at least I thought I did). 
Whatever the reason,  [this PR](https://github.com/gramps-project/gramps-web/pull/844) should solve it.

Issue #811, opened 2025-11-01
[Bug] Login form too wide on mobile


DavidMStraub:
### Bug Description

On mobile (here Chrome on Android) the login screen now fills 100% of screen width, which is ugly. We should set a max width of 90vw on the container.

![Image](https://github.com/user-attachments/assets/7365a7a5-943a-46cb-afd0-d8356d8c4dcc)

### Gramps Web Version Information

Gramps 6.0.5
Gramps Web API 3.4.1
Gramps Web Frontend 25.10.2
Gramps QL 0.4.0
Sifts 1.1.0
locale: en
multi-tree: false
task queue: true
OCR: true
chat: true

### Verification

- [x] I have searched for existing issues before creating this one

PR #820, opened 2025-11-06
Ensure login form is less than `90vw`

owenvoke:
This ensures that the login form displays correctly on mobile. Added `margin: auto` to ensure the form is centered.

Closes https://github.com/gramps-project/gramps-web/issues/811


DavidMStraub:
Thank you!!

Issue #819, opened 2025-11-05
Offering download fallback for PDF files if rendering fails


Ziesie1:
Currently, PDF files cannot be viewed on Chrome for Android.
It appears there is a bug in Chrome for Android preventing PDFs from being rendered via the <object> tag (see [this post](https://support.google.com/chrome/thread/385625592?hl=en&sjid=12039720140832423940-EU)).

This issue is therefore **not caused** by Gramps Web.

However, I suggest that it would be better to provide a download fallback for PDF files instead of showing the error message “Failed loading PDF file,” which is currently **not visible** because the toast notification is covered.

Providing a download link would improve the user experience on affected devices.
What do you think?

DavidMStraub:
True, but there is already a download button, which UI would you propose?

![Screenshot_20251105_214729_Chrome~2.jpg](https://github.com/user-attachments/assets/46fd1bc1-8f03-4f2e-8eab-dfca500ad5f6)

Ziesie1:
Ahh yes you are right. 
At first I didn't recognise it due to the huge black screen (it looks like something went wrong, so I immediately returned to the previous screen).
In my opinion it would be better if there is a message shown in the middle of the screen that something went wrong and the file can be downloaded below.

DavidMStraub:
Agreed

PR #821, opened 2025-11-07
Add PDF display error message 

Ziesie1:
This PR implements issue #819.

Changes:
- added fallback text
- Added some styling for the fallback text

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -249,5 +249,6 @@
   "Show less": "Show less",
   "Most specific position on the human Y chromosome tree": "Most specific position on the human Y chromosome tree",
   "Paternal Lineage": "Paternal Lineage",
-  "Add Y-DNA data": "Add Y-DNA data"
+  "Add Y-DNA data": "Add Y-DNA data",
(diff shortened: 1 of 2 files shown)


DavidMStraub:
Cool, thank you!

I didn't test it yet, just one thing I'm wondering: is there a case where this text is visible while the PDF is still loading or do we know that this is only displayed on systems that don't support PDFs? If it's the former, we might have to listen for some error event...

Ziesie1:
I tested it with "throttling" using the Chrome debugging tools. I didn't see the text, the PDF viewer is immediately visible. To me, it looks as if the PDF viewer itself is loading the PDF file.

I also found this line in the [HTML specifications](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-object-element): "the contents of object elements act as fallback content, used only when referenced resources can't be shown (e.g. because it returned a 404 error)."

So I would assume that everything is fine?

DavidMStraub:
Yep, sounds good, thanks for checking!

PR #822, opened 2025-11-08
Display the assigned time period in the place hierarchy view

Ziesie1:
This PR belongs to the issue #748.

It allows the user to view the corresponding dates of each direct parent place.
image

This feature based on the API changes in the place profile, [here](https://github.com/gramps-project/gramps-web-api/pull/707).

### DIFF
--- a/src/components/GrampsjsObject.js
+++ b/src/components/GrampsjsObject.js
@@ -621,7 +621,7 @@ export class GrampsjsObject extends GrampsjsAppStateMixin(LitElement) {
               <grampsjs-place-refs
                 .appState="${this.appState}"
                 .data="${this.data?.placeref_list}"
-                .profile="${this.data?.profile.direct_parent_places || []}"
+                .profile="${this.data?.profile?.direct_parent_places ?? []}"
                 ?edit="${this.edit}"
               ></grampsjs-place-refs>
             `
--- a/src/components/GrampsjsPlaceRefs.js
+++ b/src/components/GrampsjsPlaceRefs.js
@@ -26,12 +26,13 @@ export class GrampsjsPlaceRefs extends GrampsjsEditableTable {
   }
 
   row(obj, i, arr) {
-    const prof = this.places.length > i ? this.profile[i] : {}
+    const placeProf = this.places.length > i ? this.places[i].profile : {}
+    const placeRefProf = this.profile.length > i ? this.profile[i] : {}
     return html`
-      <tr @click=${() => this._handleClick(prof.place?.gramps_id)}>
-        <td>${prof.place?.name}</td>
-        <td>${prof.place?.type}</td>
-        <td>${prof.date_str}</td>
+      <tr @click=${() => this._handleClick(placeProf.gramps_id)}>
+        <td>${placeProf.name}</td>
+        <td>${placeProf.type}</td>
+        <td>${placeRefProf.date_str}</td>
         <td>
           ${this.edit
             ? this._renderActionBtns(


src/components/GrampsjsObject.js
```diff
@@ -621,6 +621,7 @@ export class GrampsjsObject extends GrampsjsAppStateMixin(LitElement) {
               <grampsjs-place-refs
                 .appState="${this.appState}"
                 .data="${this.data?.placeref_list}"
+                .profile="${this.data?.profile.direct_parent_places || []}"
```
DavidMStraub:
```suggestion
                .profile="${this.data?.profile?.direct_parent_places ?? []}"
```

src/components/GrampsjsPlaceRefs.js
```diff
@@ -20,16 +20,18 @@ export class GrampsjsPlaceRefs extends GrampsjsEditableTable {
...
     this.objType = 'Place'
-    this._columns = ['Name', 'Type', '']
+    this._columns = ['Name', 'Type', 'Date', '']
     this.dialogContent = ''
+    this.profile = []
   }
 
   row(obj, i, arr) {
-    const prof = this.places.length > i ? this.places[i].profile : {}
+    const prof = this.places.length > i ? this.profile[i] : {}
```
DavidMStraub: This breaks backward compatibility, because now if the backend is not the latest version (i.e. the currently unreleased one), not even name or type will be displayed anymore.

Please leave the existing code as is and just add the date string *if* the backend provides it.

src/components/GrampsjsPlaceRefs.js
```diff
@@ -20,16 +20,18 @@ export class GrampsjsPlaceRefs extends GrampsjsEditableTable {
     super()
     this.places = []
     this.objType = 'Place'
-    this._columns = ['Name', 'Type', '']
+    this._columns = ['Name', 'Type', 'Date', '']
     this.dialogContent = ''
+    this.profile = []
```
DavidMStraub, line 25: I find this parameter name too generic - this is only the `direct_parent_places` part of the profile.

DavidMStraub:
Thanks for this!

Added a couple of comments.

src/components/GrampsjsObject.js
Ziesie1: Why do you suggest ?? instead of ||.
I agree that in a worst case scenario only, a null value will occur, so ?? is sufficient.

On the other hand, in the most other cases (for example [alternate_place_names](https://github.com/gramps-project/gramps-web/blob/cffac3b239886cba4df6783185cdf1c319ab2c34/src/components/GrampsjsObject.js#L613)) the  || operator is used.
Then it should be adjusted consistently, shouldn't it?

DavidMStraub: My main point here is the first question mark, this one is important.

Replacing || by ?? is just something I did since I was already typing the line. I don't think it makes a difference here in practice, but I find it less misleading. It says explicitly, if the LHS is `null` or `undefined`, use `[]`. `||` is more appropriate in places where the LHS can also be something else that's falsey, such as an empty string, and we want to replace that by a default value.

It's true that the overall project currently is not consistent in this respect - this is because a lot of the code was initially written by a Python guy who probably was neither aware of an empty array being trueish in JS, nor of the existence of `??`.

src/components/GrampsjsPlaceRefs.js
Ziesie1: I tried to stick to the implementation in [grampsjs-place-names](https://github.com/gramps-project/gramps-web/blob/cffac3b239886cba4df6783185cdf1c319ab2c34/src/components/GrampsjsObject.js#L613).
Here we only use the `alternate_place_names` from the place profile and store them in a `profile` variable for the `grampsjs-place-names`.

According to my mental model, the variable `alternate_place_names` represents the profile for `grampsjs-place-names`. That's why I think it is okay to just call it “profile”.
The same applies to `direct_parent_places` and `grampsjs-place-refs`. `direct_parent_places` represents the profile for the place references. For this reason, I originally proposed the API key `direct_parent_places` “place_references.” 

However, if you prefer an explicit name, I can rename the variable. Then to `direct_parent_places`?

DavidMStraub: No, you are right actually! Let's leave it.

Ziesie1:
Thanks for your feedback. PR is now backward compatible (also tested).
I only renamed the variable `profile` to `placeProf` because this information is collected about the profile of each individual place. And, of course, to achieve a clearer separation from “placeRefProf.”

PR #826, opened 2025-11-12
feat: add Help link to navigation dropdown

### LINKED ISSUE #501: Enhancement - Put a Link to a User Guide / Manual in the top nav bar.
Put a Icon / Link to the User Manual.  Recommend it go to a "User Guide" section, targeted for laymen end-users rather than Developers or Administrators, which will have NO problems learning the app.  This is targeted for family members who may only use it part time and need to have ready access to a Manual with examples and such.  t

image

owenvoke:
Not 100% sure if this is what's wanted, or whether it should be dynamic based on the specific page that is being accessed.

Or if it should be a link directly to [the actual user guide section](https://grampsweb.org/user-guide), however it's maybe a little sparse on first viewing.

Closes #501

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -250,5 +250,9 @@
   "Most specific position on the human Y chromosome tree": "Most specific position on the human Y chromosome tree",
   "Paternal Lineage": "Paternal Lineage",
   "Add Y-DNA data": "Add Y-DNA data",
-  "Unfortunately, your browser does not support the display of PDF files. To view the file anyway, you can download it using the button below.": "Unfortunately, your browser does not support the display of PDF files. To view the file anyway, you can download it using the button below."
+  "Unfortunately, your browser does not support the display of PDF files. To view the file anyway, you can download it using the button below.": "Unfortunately, your browser does not support the display of PDF files. To view the file anyway, you can download it using the button below.",
+  "Help": "Help",
+  "User Documentation": "User Documentation",
+  "Administrator Documentation": "Administrator Documentation",
+  "Forum": "Forum"
 }
--- a/src/views/GrampsjsViewHelp.js
+++ b/src/views/GrampsjsViewHelp.js
@@ -1,40 +1,64 @@
-import {html} from 'lit'
+import {css, html} from 'lit'
 
+import {mdiOpenInNew} from '@mdi/js'
 import {GrampsjsView} from './GrampsjsView.js'
 
 export class GrampsjsViewHelp extends GrampsjsView {
+  static get styles() {
+    return [
+      super.styles,
+      css`
+        .button-container {
+          display: flex;
+          gap: 0.5em;
+          margin-top: 1em;
+          margin-bottom: 0.5em;
+          --button-height: 48px;
+        }
+      `,
+    ]
+  }
+
   renderContent() {
     return html`
       <h2>${this._('Help')}</h2>
 
       <div>
-        <mwc-button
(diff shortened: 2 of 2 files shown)


DavidMStraub:
Thanks for implementing this!

~~I think it's a good idea, but I'd wait a bit before deciding whether to merge it, as #501 would be even more helpful and I don't think we need both.~~

Edit: I'm sorry, it's too early in the morning. You wrote that the PR fixes #501.

DavidMStraub:
LOL, re-reading this other issue which was written before the settings button was a dropdown, I wrote

>  I like Nextcloud's solution of making the app bar user button a dropdown that links to settings, help, etc.

which is exactly what you did :blush:

So, definitely a good idea :wink:

DavidMStraub:
In the case of Nextcloud, clicking help in the Dropdown opens a very simple local page:

grafik

owenvoke:
I guess it could have a new local page within the Gramps Web frontend (similar to that NextCloud one) with links to:
- [User Documentation](https://www.grampsweb.org/user-guide)
- [Administrator Documentation](https://www.grampsweb.org/administration/admin)
- [Forum](https://gramps.discourse.group/c/gramps-web)

Or can leave it as-is, sorry I'm not sure if the screenshot was a request or just showing what the NextCloud one looks like.

DavidMStraub:
> Or can leave it as-is, sorry I'm not sure if the screenshot was a request or just showing what the NextCloud one looks like.

Not sure either :laughing: What do you think, would it be useful?

What I like about it is that, otherwise, a user might be surprised that one of the links in the dropdown takes them out of the app while the others don't. So at the very least, we would need a ↗️ like symbol. It would also solve the question where exactly we link (user docs, admin docs or landing page).

owenvoke:
I think a separate page would make sense. I'll mark this PR as draft and work on that later. 👍🏻

owenvoke:
@DavidMStraub, I've now updated this to use a dedicated page. Here's a preview of how it looks:

Screenshot 2025-11-12 at 12 08 18

docker-compose.yml
```diff
@@ -1,8 +1,6 @@
 # docker-compose file for frontend development
 # do not use in production!
-# https://www.grampsweb.org/dev-frontend/setup/
-
-version: "3.8"
+# https://www.grampsweb.org/development/frontend/setup/
```
owenvoke, line 3: Just thought I'd fix this link, and removed the `version` as it's no longer a valid parameter.

DavidMStraub:
Beautiful, thank you!

DavidMStraub:
Beautiful!

Since we're in the process of migration from `mwc-` to `md-` components, would you mind using the following buttons instead:


```javascript
import { mdiOpenInNew } from '@mdi/js'

<md-outlined-button
  href="https://gramps.discourse.group/c/gramps-web"
  target="_blank"
>
    ${this._('Forum')}
    <grampsjs-icon
      .path="${mdiOpenInNew}"
      slot="icon"
      color="var(--mdc-theme-primary)"
    ></grampsjs-icon>
</md-outlined-button>
```

etc.

Also, can you please add the three button texts to `en.json`?

Thank you!

owenvoke:
@DavidMStraub, I've updated to use those instead. 👍🏻

Screenshot 2025-11-13 at 08 18 08

DavidMStraub:
Awesome! :rocket:

Issue #388, opened 2024-02-10
Emergency enhancement for 17 Feb 2024 Digital Services Act (DSA) compliance?


emyoulation:
Does Gramps Web need a new enhancement to comply with the DSA requirement requirement for online platforms sites to start making it easy to report illegal content on the platforms?

> "Tobias Eigen, post:1, [Discourse topic:294866](https://meta.discourse.org/t/new-its-illegal-reason-when-flagging-posts/294866)"
> The reason for this change is the upcoming 17 February, 2024 Digital Services Act (DSA) requirement for online platforms like Discourse sites to start making it easy to report illegal content on the platforms.

DavidMStraub:
Good question. Generally, a tree owner will be responsible for moderating complaints about illegal content. In most cases, the users having access to a tree will be people who personally know each other, so this will not be relevant, but it is conceivable that an individual or organizations makes a tree accessible to a large number of users and this might become relevant.

What we could do is to make the "researcher" details visible in the frontend. The information is anyway already provided by the API and it can hold information about the address, e-mail, and phone contacts of the tree owner. We could add a comment to the documentation that we advise tree owners to enter this information for legal reasons to allow tree users to contact them in the case of suspsected illegal activities.

PR #827, opened 2025-11-13
feat: add Researcher Information tab

owenvoke:
This adds a new "Researcher Information" page in the settings tabs.

Here's a preview:

Screenshot 2025-11-13 at 08 48 16

It uses the details from the app state. I wasn't sure if you'd rather hide the sections if they are empty strings, or add a `-` to indicate that it's empty. 🤷🏻

I wasn't sure if `addr` usually contains the full address, or just part of it. If it's just part of it, perhaps I'll need to concatenate the different segments from `addr`, `locality`, `country`, etc.

I also added configurability through the `frontendConfig` to hide this with `hideResearcherDetails: false` 👍🏻

Closes https://github.com/gramps-project/gramps-web/issues/388


DavidMStraub:
Thanks for this!

Comments:

- We have two ways to get translation strings: through the backend via Gramps core or through dedicated strings for the frontend. `en.json` is only for the latter, but all strings needed here are already translated in Gramps core, so there is no need to add the to `en.json`. Instead, they should be added to the list in `strings.js`. Some of them, like "E-mail" are already there; for the phone number, it should be "Phone". You can check [one of the .po files in Gramps core](https://raw.githubusercontent.com/gramps-project/gramps/master/po/de.po) to check which strings exist.

- The address is just a string, but since `Researcher` [inherits from](https://github.com/gramps-project/gramps/blob/master/gramps/gen/lib/researcher.py#L38) `LocationBase`, there is a large number of additional possible parameters. I would add exactly the ones that also Gramps Desktop displays and allows changing, see here:

https://github.com/gramps-project/gramps/blob/master/gramps/gui/configure.py#L689-L754

There you can also see the extract strings to be added to `strings.js`.

The way I understand it, "Address" is supposed to be the street name/number, since `LocationBase.street` is not used in this view.

Since, in the end, the research details are name, e-mail, phone, and address pieces, I think it would be prettier to display it more like an address book page, what do you think?

### John Doe
**E-Mail:** [john.doe@example.com](mailto:john.doe@example.com)
**Phone:** +1 234 56789

**Address:**
...

owenvoke:
Sounds good to me! Thanks for the information and feedback, I'll update this. 👍🏻

How's this?

Screenshot 2025-11-13 at 11 47 48

DavidMStraub:
Perfect! :partying_face:

PR #829, opened 2025-11-14
Form new event: fire event when setting default value

### LINKED ISSUE #828: [Bug] Events default role won't be saved
### Bug Description

When creating a new event for an existing family, the “Family” role is preselected in the dialog box by default:

Image

If you leave the preselection unchanged and “save” the event, the “Family” role will not be saved correctly. Instead, the role will change to “Primary”:

Image

This hidden error is really annoying because you don't see it at first and it causes the title of a marriage event to be empty in several events (the “Family” role is required here):

Image

Gramps 6.0.5
Gramps Web API 3.4.1
Gramps Web Frontend 25.10.2
Gramps QL 0.4.0
Sifts 1.1.0
locale: en
multi-tree: false
task queue: true
OCR: true
chat: false

### Verification

DavidMStraub:
Fixes #828.

### DIFF
--- a/src/components/GrampsjsFormSelectType.js
+++ b/src/components/GrampsjsFormSelectType.js
@@ -79,7 +79,7 @@ class GrampsjsFormSelectType extends GrampsjsAppStateMixin(LitElement) {
         !types.includes(this.value) &&
         types.includes(this.defaultValue)
       ) {
-        this.value = this.defaultValue
+        this.#setValue(this.defaultValue)
       }
     }
   }


m-breitbach:
Thanks for fixing. I have currently no working dev setup to test, but could there be side effects in any of the use cases? Like evaluation hints showing up immediately, or immediate state changes? I have seen nothing specifically in the code, just wanted to raise the possibility. (I remember that when I worked on the behavior of that whole element, everything broke really easiliy and was hard to get right.)

DavidMStraub:
Your're right about it breaking easily.

I think the main effect is that the default role will always be part of the payload (rather than staying empty and being filled in by the backend when the select is not modified) but I think that's ok. Since the only change is updating the parent form's data when the type value is changed in all cases, it should prevent any mismatch between the form and the component.

Issue #57, opened 2021-11-20
Source: allow editing author and repository reference


DavidMStraub:

jittymolmathew92:
As per the discussion with @DavidMStraub , the abbreviation, author and public information made editable for a Source

DavidMStraub:
Issue status:

- [x] Edit Author
- [ ] Edit repository reference

DavidMStraub:
Fixed by #925!

PR #832, opened 2025-11-16
Source: allow editing Abbreviation, Author and Pub info

jittymolmathew92:
Please review he code against feature - https://github.com/gramps-project/gramps-web/issues/57

### DIFF
--- a/src/components/GrampsjsFormEditSourceData.js
+++ b/src/components/GrampsjsFormEditSourceData.js
@@ -1,5 +1,5 @@
 /*
-Form for adding a new event reference
+Form for updating the propertoies of Source - Abbrevation, Author & Pub Info
 */
 
 import {html} from 'lit'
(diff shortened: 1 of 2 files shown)


src/components/GrampsjsFormEditSourceData.js
```diff
@@ -0,0 +1,61 @@
+/*
+Form for adding a new event reference
```
DavidMStraub: The comment needs to be adapted

src/components/GrampsjsSource.js
```diff
@@ -101,6 +111,23 @@ export class GrampsjsSource extends GrampsjsObject {
...
+
+  _handleSourceData() {
+    const data = {
+      abbrev: this.data?.abbrev,
+      author: this.data?.author,
+      pubinfo: this.data?.pubinfo,
+    }
+    this.dialogContent = html`
+      <grampsjs-form-edit-source-data
+        @object:save="${this._handleSaveTitle}"
```
DavidMStraub: Ah, it's quite nice that the same existing callback can be used that exists already for the title, but we should then rename it (`_handleSaveProps` or so).

src/components/GrampsjsFormEditSourceData.js
```diff
@@ -0,0 +1,61 @@
...
+class GrampsjsFormEditSourceData extends GrampsjsObjectForm {
+  static get properties() {
+    return {
+      prop: {type: Object},
+    }
+  }
+
+  constructor() {
+    super()
+    this.prop = {abbrev: 'abbrev', author: 'author', pubinfo: 'pubinfo'}
```
DavidMStraub: If I remember correctly, the `GrampsjsObjectForm` already has an existing empty `this.data` object that can be used for this purpose instead of `this.prop` - then we can completely remove the `get properties` above because the one in the parent class is enough.

DavidMStraub:
Great, thanks! Just added 3 minor comments.

src/components/GrampsjsFormEditSourceData.js
jittymolmathew92: done

src/components/GrampsjsSource.js
jittymolmathew92: Done

src/components/GrampsjsFormEditSourceData.js
jittymolmathew92: Done, please check is this what really expected

DavidMStraub: Looks good I think! If it works :slightly_smiling_face:

jittymolmathew92: yeah it is working in my local, I tested it

DavidMStraub:
Works perfectly, thanks!

Issue #831, opened 2025-11-15
Simplify filtering for the global search


Ziesie1:
### Problem
In most cases (around 90% of the time), I am searching for a person and do not want to see all media or other related objects.
Currently, filtering requires two clicks every time:

- Click “All” to deselect everything,
- Then select “Persons” to filter only by persons.

This extra step is disruptive and slows down common workflows.

### Proposed Solution

- No filter selected by default: When opening the search, no filter category should be pre-selected.
- If no filter is active, show all results: Selecting zero filters should behave the same as “show everything.”
- User can activate one or more filters: Once the user selects filters, only the corresponding item types should be displayed.
- Remove the “All” button: With the logic above, this button becomes unnecessary.

<img width="4988" height="2492" alt="Image" src="https://github.com/user-attachments/assets/e41128c8-14c1-41a7-9a36-78e73d801c41" />

<img width="4980" height="1760" alt="Image" src="https://github.com/user-attachments/assets/03d519f7-6eb5-40f9-ab49-170270f81e1d" />

This solution allows filtering with just one click.

DavidMStraub:
You're absolutely right.

In fact, this filter was originally inspired by the LinkedIn search filter, and what you describe is exactly the way it behaves.

Ziesie1:
Great, then I'll get to work on it in the next few days 🙃.

PR #834, opened 2025-11-16
Simplify global search filters (#831)

Ziesie1:
Fix issue #831.

This pull request refactors the object type filter logic in the `GrampsjsViewSearch` component to improve usability. The main change is that no object types are selected by default, and the "All" toggle option has been removed. The filtering logic now treats the case where no filters are selected as equivalent to searching for all object types, rather than showing a "Select at least one object type" message.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -64,8 +64,6 @@
   "Continue with %s": "Continue with %s",
   "Signing in...": "Signing in...",
   "Account Source": "Account Source",
-  "All": "All",
-  "Select at least one object type": "Select at least one object type",
   "filter": "filter",
   "Properties": "Properties",
   "unconfirmed": "unconfirmed",


Ziesie1:
I just remembered that there are still some unused texts ("all" & "Select at least one object Tag"). Should these be deleted?
Is it sufficient to delete them from the english translations only, or do they need to be removed from all languages manually?

DavidMStraub:
> I just remembered that there are still some unused texts ("all" & "Select at least one object Tag"). Should these be deleted?

Good point, yes!

> Is it sufficient to delete them from the english translations only, or do they need to be removed from all languages manually?

English is sufficient. Weblate should take care of the rest.

Ziesie1:
Texts have been removed.

DavidMStraub:
Works great!

Issue #569, opened 2025-01-13
Create persons from family view


kenayagi:
tl;dr: I would like to create persons on the fly while creating a family.

I find quite annoying to have to firstly create persons and link them in families only after.
The only choice is to add an **existing** person:

![Image](https://github.com/user-attachments/assets/79f0731f-1021-4a65-92e6-88ccad421c8c)

On Gramps desktop it's instead quite easy by simply clicking the PLUS sign:

![Image](https://github.com/user-attachments/assets/a6dfb474-0a2e-408b-bf77-c1005ffedbf9)

Is it maybe possible to implement a similar behaviour even on Gramps web?
Or am I simply doing something wrong?

(Request similar to #227 )

kenayagi:
Thanks @DavidMStraub 😊👌

(and double thanks for this AWESOME software! 🔝)

DavidMStraub:
Implemented in #841.

PR #841, opened 2025-11-22
Create persons from family view

mschuepbach:
#569

I made an attempt at implementing this issue but i am having some problems.

With the current code, the select object lists aren't updating after creating a new person. I was able to fix this in the edit family view by changing `firstUpdated()` in `GrampsjsFormSelectObjectList.js` to `updated()`. But that causes an infinite fetch loop in the new family view when selecting a person and the browser freezes.

### DIFF
--- a/src/components/GrampsjsFamily.js
+++ b/src/components/GrampsjsFamily.js
@@ -44,7 +44,7 @@ export class GrampsjsFamily extends GrampsjsObject {
   renderProfile() {
     return html`
       <h2>${this._renderTitle()}</h2>
-      ${this._renderFather()} ${this._renderMother()}
+      ${this._renderParent('father')} ${this._renderParent('mother')}
       <div class="relationship-type">
         ${this._renderRelType()} ${this._renderMarriage()}
         ${this._renderDivorce()}
@@ -81,20 +81,6 @@ export class GrampsjsFamily extends GrampsjsObject {
     `
   }
 
-  _renderFather() {
-    return html`
-      <h4 class="label">${this._('Father')}</h4>
-      ${this._renderParent('father')}
-    `
-  }
-
-  _renderMother() {
-    return html`
-      <h4 class="label">${this._('Mother')}</h4>
-      ${this._renderParent('mother')}
-    `
-  }
-
   _renderParent(parent) {
     const profile = this.data?.profile[parent]
     const hasProfile = Object.keys(profile ?? {}).length > 0
@@ -102,7 +88,7 @@ export class GrampsjsFamily extends GrampsjsObject {
     return html`
       <div class="parent">
         <div class="items-center">
-          ${!this.edit || hasProfile ? renderPerson(profile || {}) : ''}
+          ${renderPerson(profile || {})}
           ${this.edit && hasProfile
             ? html`
                 <mwc-icon-button


DavidMStraub:
Can you please describe in more detail (ideally screenshots) what it is that's not updating?

mschuepbach:
Sure. I added the "add a new person" buttons in the family edit view and moved the "select an existing person" select out from the dialog.

image

"add a new person" opens a dialog for the new person and saving adds the new person to the family. But this is not really noticeable. Only the title of the family updates with the name. After stopping editing and edit again it displays the new person as the selected one.

image

DavidMStraub:
I find the UI for editing the parents too radically different from what we do in the rest of the app.

I would suggest the following:

- if a parent is missing, display an "add" and "link" button instead of the name. They can behave just as they do for adding a child.
- If a parent is not missing, we can simply add an unlink (or delete) button next to the name (like the one in your screenshot, but in blue)

I suspect your problem with the select object lists will disappear if they are in the object form dialogs as for children rather than in the main view.

mschuepbach:
And what about the new family view? I think it would be nice to also be able to create new people from there. Or should I leave that away for now? I think I would have to replace the select object lists and use an editable table for the children.

DavidMStraub:
> And what about the new family view? I think it would be nice to also be able to create new people from there. Or should I leave that away for now?

Yes, please let's leave the "new family" view out of scope of this PR (and also out of scope of #569). The "new family" view needs a different design without popup dialogs, but we can discuss that in a new issue first.

mschuepbach:
Should i move the edit button for the relationship type to the right of the text or keep it where it is?

image

DavidMStraub:
Looks great

> Should i move the edit button for the relationship type to the right of the text or keep it where it is?

If it's to the right of the marriage event I think it's a bit misleading (because we can't edit the marriage event from  there). Perhaps moving it a few pixels up would make it less detached. Or we move it between the relationship type and the marriage event, not sure if that would look good.

mschuepbach:
Or don't render the marriage and divorce events when editing?

image

mschuepbach:
This is how it looks with the button on a separate line but closer.

image

DavidMStraub:
I'm ok with this version.

src/components/GrampsjsFamily.js
```diff
@@ -67,11 +82,53 @@ export class GrampsjsFamily extends GrampsjsObject {
   }
 
   _renderFather() {
-    return html` <p>${renderPerson(this.data?.profile?.father || {})}</p> `
+    return html`
+      <h4 class="label">${this._('Father')}</h4>
```
DavidMStraub: So far, since we didn't show the string "Father" the family view worked well for same-sex couples as well. It would be nice if we could handle that here as well.

Gramps desktop shows "Father/partner1" (although I am puzzled about the notation without a space); in my opinion it would be cleaner if we default to Father/Mother and switch to "Partner 1"/"Partner 2" if the partners have the same gender.

mschuepbach: Father/Mother also doesn't make sense if they don't have any children. Can I just remove the labels again to avoid that issue for now? I think it should be clear what the buttons are for if I add enough margins. And then this can be addressed in another PR.

DavidMStraub: You're right.

If that solution is ok for you, let's go with that.

DavidMStraub:
I tested it and it works great, excellent work!

PR #844, opened 2025-11-24
fix color for mwc-icon-button in meta slot

Ziesie1:
This PR fixes the issue [here](https://github.com/gramps-project/gramps-web/pull/818).
Now light and darkmode should work properly.


DavidMStraub:
Cool, thanks!

PR #856, opened 2025-12-02
Customize search icon color for dark mode support

Ziesie1:
Currently, the search icon in the search bar is not displayed correctly in dark mode:
image

This PR fixes this issue, by modifying the icon child inside of the `<mwc-textfield>` component:
image

To be honest, I don't like this solution, but as far as I know, this is the only way to change the icon color of an `<mwc-textfield>` element. An alternative approach would be to remove the icon completely or use a different component.


DavidMStraub:
Sorry for ignoring this, indeed it's quite ugly - I'll first replace mwc- by md-elements.

PR #862, opened 2025-12-07
External search

jittymolmathew92:


DavidMStraub:
Nice work, the two search engines already work well! Let's also add CompGen meta search to the initial options, the endpoint is

`https://meta.genealogy.net/search?lastname=Straub&place=M%C3%BCnchen` (both or one of the parameters can be provided)

Non-source-code comments:

- when all 3 buttons are shown, the gaps between them are not the same. Let's also make sure they look nicely on small (mobile) screens, probably using flexbox

grafik

- Let's add an "external link" icon behind the provider links to set the user expectation that this will open in a new tab. You can use `grampsjs-icon` and `mdiOpenInNew` (like in `GrampsjsViewHelp.js`).
- We also need to an an explanatory sentence to the dialog - something like "Select a web service to search for the person's details. Some of the services require prior registration." - I'm open to suggestions how to formulate it well.

src/components/GrampsjsForMorePersonDetails.js
```diff
@@ -0,0 +1,105 @@
...
+
+import {html} from 'lit'
+import '@material/mwc-icon'
+import '@material/mwc-list'
+import '@material/mwc-list/mwc-list-item'
+import './GrampsjsTooltip.js'
+import './GrampsjsFormSelectType.js'
+import {GrampsjsObjectForm} from './GrampsjsObjectForm.js'
+
+class GrampsjsForMorePersonDetails extends GrampsjsObjectForm {
```
DavidMStraub: `For` should be `Form`, no?

Since this form is only used for external search, what about `GrampsjsFormExternalSearch`?

src/components/GrampsjsForMorePersonDetails.js
```diff
@@ -0,0 +1,105 @@
...
+        </md-list>
+      </div>
+    `
+  }
+
+  _handleClick(baseUrl, key) {
+    let url = ''
+    if (baseUrl && key) {
+      switch (key) {
+        case 'familySearch':
```
DavidMStraub: To keep this code maintainable when adding more providers, I suggest to use the `searchWebUrls` object for the whole URL, not just the base URL, using `{{...}}` placeholders, eg. `..?q.anyPlace={{place_name}}...`

The `handleClick` function then becomes provider indepdent and only needs to replace a number of predefined placeholders with the correct value.

src/components/GrampsjsObjectForm.js
```diff
@@ -131,11 +135,15 @@ export class GrampsjsObjectForm extends GrampsjsAppStateMixin(LitElement) {
           ${this.dialogIsOpen ? this.renderForm() : ''}
         </div>
         <div slot="actions">
-          <md-text-button class="edit" @click="${this._handleDialogCancel}">
+          <md-text-button
+            class="edit ${this.showCancelButton === false ? 'hide' : ''}"
```
DavidMStraub: Simpler:  `this.showSaveButton ? 'hide' : ''`.

src/components/GrampsjsObjectForm.js
```diff
@@ -131,11 +135,15 @@ export class GrampsjsObjectForm extends GrampsjsAppStateMixin(LitElement) {
...
+            @click="${this._handleDialogCancel}"
+            ${this.showButton}
+          >
             ${this._('Cancel')}
           </md-text-button>
           <md-filled-button
-            class="edit"
+            class="edit ${this.showSaveButton === false ? 'hide' : ''}"
```
DavidMStraub: Same here

src/components/GrampsjsObjectForm.js
```diff
@@ -69,6 +71,8 @@ export class GrampsjsObjectForm extends GrampsjsAppStateMixin(LitElement) {
     this.isFormValid = false
     this.new = false
     this.dialogTitle = ''
+    this.showSaveButton = this.showSaveButton || true
```
DavidMStraub:
```
this.showSaveButton = true
```

This is the default value, you can then override it in child classes (like the external search form) in the constructor.

src/components/GrampsjsObjectForm.js
```diff
@@ -69,6 +71,8 @@ export class GrampsjsObjectForm extends GrampsjsAppStateMixin(LitElement) {
...
+    this.showCancelButton = this.showCancelButton || true
```
DavidMStraub: Same here

DavidMStraub:
Let's please also make the external search button the last of the 3 buttons.

src/components/GrampsjsPerson.js
```diff
@@ -119,15 +117,26 @@ export class GrampsjsPerson extends GrampsjsObject {
...
+          ${this._('Show in tree')}
+          <grampsjs-icon
+            path="${mdiFamilyTree}"
+            color="var(--mdc-theme-primary)"
+            slot="icon"
+          >
+          </grampsjs-icon>
+        </md-outlined-button>
+        <md-outlined-button @click="${this._handleMoreDetailsClick}">
+          ${this._('External Search')}
```
DavidMStraub: Eventually, we also need to add this to `en.json`.

I'm still unsure whether "External Search" or "Web Search" is better - the first is a bit clearer and the second a bit shorter. What do you think?

In case we use "Web Search", we don't need to add a new string, as it's already present in `strings.js` (backend strings).

src/components/GrampsjsForMorePersonDetails.js
jittymolmathew92: Done

DavidMStraub: But not pushed yet, right?

src/components/GrampsjsObjectForm.js
jittymolmathew92: Done

src/components/GrampsjsObjectForm.js
jittymolmathew92: Done

src/components/GrampsjsObjectForm.js
jittymolmathew92: Done

src/components/GrampsjsObjectForm.js
jittymolmathew92: Done

src/components/GrampsjsPerson.js
jittymolmathew92: I think its better to use the External Search, it is more clear

src/components/GrampsjsPerson.js
```diff
@@ -162,6 +175,32 @@ export class GrampsjsPerson extends GrampsjsObject {
...
+    }
+    this.dialogContent = html`
+      <div>
+        <grampsjs-form-external-search
+          @object:cancel=${this._handleCancelDialog}
+          .appState="${this.appState}"
+          .data=${data}
+          .dialogTitle=${this._('External Search')}
+          .showSaveButton=${false}
+          .showCancelButton=${true}
```
DavidMStraub: It's ok like this for now (so no need to change it before merging), but FYI usually it's better, when using boolean properties, if they default to `false` (this could be achieved e.g. by using `hideSaveButton`, `hideCancelButton`), see https://lit.dev/docs/components/properties/#boolean-attributes. Then one can set the boolean to true with pure HTML.

lang/en.json
```diff
@@ -255,5 +255,7 @@
   "Quick Access": "Quick Access",
   "Documentation": "Documentation",
   "Community": "Community",
-  "Researcher Information": "Researcher Information"
+  "Researcher Information": "Researcher Information",
+  "External Search": "External Search",
+  "Select a web service to search for the person details. Some of the services require prior registration": "Select a web service to search for the person details. Some of the services require prior registration"
```
DavidMStraub: Let's please change it to "Select a web service to search for the person’s details. Some services require prior registration."

src/components/GrampsjsFormExternalSearch.js
```diff
@@ -0,0 +1,110 @@
...
+      },
+      {
+        key: 'ancestry',
+        value: 'Ancestry',
+        baseUrl: `https://www.ancestry.com/search`,
+        params: `?name=${this.data.name_given}_${this.data.name_surname}&event=${this.data.place_name}&searchMode=advanced`,
+      },
+      {
+        key: 'genealogy',
+        value: 'Genealogy',
```
DavidMStraub: Let's call this `CompGen`.

DavidMStraub: And can you please move it to the top (first option)?

DavidMStraub:
Looks great now!

Can you please move the array of search engines and URLs output of the component class into a constant object with placeholders (like `...{{surname}}...` and then add a method `this._buildSearchUrl(searchEngineObject)` which builds the search URL from one of the items in the array? This would make the `render` function much cleaner and separate the logic from the metadata.

Concerning the UI, the final thing I'm wondering is how we can make it more visible before a user clicks which search engine requires prior registration. Perhaps we can add a little icon after each search engine

Something like

- https://pictogrammers.com/library/mdi/icon/book-open-blank-variant-outline/ or https://pictogrammers.com/library/mdi/icon/earth/ for open/free
- https://pictogrammers.com/library/mdi/icon/shield-account/ for registration required
- https://pictogrammers.com/library/mdi/icon/account-credit-card/ or https://pictogrammers.com/library/mdi/icon/account-cash/ for subscription required

What do you think?

src/components/GrampsjsPerson.js
jittymolmathew92: ok will do that change along with the icon changes

lang/en.json
jittymolmathew92: So I am afraid that the 's makes any issue in lang file like any special character issue. Because for key also we are using same entire text. And I am wondering why we can't use a specific unique small key for the description string in lang file?

src/components/GrampsjsFormExternalSearch.js
jittymolmathew92: Sure

jittymolmathew92:
As per my suggestion, we can place the criteria icons left before the search engines instead of open-in-new icon, and the design look consistent with the sidebar main menu. I opt the following icons for each criteria, please help to confirm

Open/free - https://pictogrammers.com/library/mdi/icon/earth/
Registration required - http://pictogrammers.com/library/mdi/icon/shield-lock-outline/
Paid subscription required - https://pictogrammers.com/library/mdi/icon/cash-lock/

lang/en.json
DavidMStraub: > So I am afraid that the 's makes any issue in lang file like any special character issue.

You're right - the way it's done, any change in the original string will invalidate all translations. That's precisely why we should be very picky and precise about the exact string *before* merging the PR, as translators will get to work on it after it's merged.

> And I am wondering why we can't use a specific unique small key for the description string in lang file?

You are absolutely right that this would have big advantages, and it's often done like that. One practical reason why I initially implemented it like that (without spending too much thought on it) is because that's the way it's usually done in Python, and also the way it's done in Gramps core. One possible disadvantage of using short keys is that the source code becomes less readable, because one has too look up the language file to understand what text is where.

DavidMStraub:
Yes, the icons look good! 

I think it's more intuitive if we leave the "open in new" icons as they are and add a (smaller lighter) icon to the right of the name (inline) - the "open in new" is the action, while the other icon is more like metadata about the service.

DavidMStraub:
> Can you please move the array of search engines and URLs output of the component class into a constant object with placeholders (like ...{{surname}}... and then add a method this._buildSearchUrl(searchEngineObject) which builds the search URL from one of the items in the array? This would make the render function much cleaner and separate the logic from the metadata.

Implemented in 2e841b678a11208bc1840b3b6e2e45a3025f8a4c.

DavidMStraub:
I added a legend to the icons and changed the wording a bit, this is what it looks like now:

grafik

PR #863, opened 2025-12-09
Update zh_CN.json for missiong translation

bawanag:


DavidMStraub:
Hi,

Thanks for the contribution, but unfortunately translations cannot be contributed via Github but only via Weblate. It would be great if you could add your changes there, thanks!

See https://www.grampsweb.org/zh/development/frontend/translation/

PR #889, opened 2025-12-28
Add missing translations for reports

V-donck:
All strings used in following reports:
* Ancestor Tree
* Calendar
* Descendant Tree
* Family Descendant Tree
* Fan Chart
* Statistics Chart
* Timeline Chart
* Ahnentafel Report
* Birthday and Anniversary Report
* Decendant Report
* Detailed Ancestral Report
* Detailed Descendant Report
* End of Line Report
* Family Group Report
* Complete individual report
* Kinship report
* Tag report
* Place report
* Database summary report
* Records report
* Family lines graph
* hourglass graph
* relationship graph

Except some exceptions: e.g. 
- timeline>Determines what people are included in the report>Descendants of Garner von Zieliński, Lewis Anderson Sr
- Descendant Tree>The format and language for dates, with examples> Mon Day, Year (<Current date in that format>)

So all 'normal' translations are added for the default reports.


DavidMStraub:
This is much needed, thanks a lot!

Can you please add a bit more detail whether you systematically added all missing strings and checked whether the new strings in the list exist in the backend?

V-donck:
I indeed systematically added all missing strings for the above mentioned reports (except some exceptions like the one mentioned , I could not find quickly how I could add those). 

Most of the strings are available and translated, but some are not. These are mostly options on a picklist, but also some help texts under the input fields.
- some are available in the .po file:
e.g. timeline chart>how to handle living people> included, and all data.
The translations api (nl) returns:     {
        "original": "Included, and all data",
        "translation": "Included, and all data"
    },
https://github.com/gramps-project/gramps/blob/master/po/nl.po#L10605

- some are not available:
    - Ancestor Tree> The number of generations of empty boxes that will be displayed> 2 generations of empty boxes....
    - 'Any text'
    - 'A number'
    - 'A list of text values. Each entry in the list represents one line of text.'
    - ....


I will try to add the strings for the other reports as well, but I wanted to start with this already to make sure it is a correct way of doing it.

DavidMStraub:
Ok great.

Do you want to add the other strings in the same PR or in a separate one?

V-donck:
I have added all strings for all default reports now. So I think this PR can be merged. I will try to fix the exceptions in another PR (when I have time)

Issue #909, opened 2026-01-11
[Bug] Drop Down field shows previous value in some "New" views.


Shaker1978:
### Bug Description

Choose "New place".
Give it a name and select a different type than "Unknown".
Save this palce.
Choose "New place" again.

The type drop down shows the last value, but when you save the place it is not stored, unless you have changed the drop down value.

### Gramps Web Version Information

Gramps 6.0.6
Gramps Web API 3.6.0
Gramps Web Frontend 25.12.0
Gramps QL 0.4.0
Sifts 1.1.1
locale: en
multi-tree: false
task queue: false
OCR: true
chat: false

### Verification

- [x] I have searched for existing issues before creating this one

DavidMStraub:
Fixed by #910.

PR #910, opened 2026-01-11
Bugfix for #909: GrampsjsFormSelectType in New-Dialogs did not reset its previous value

Shaker1978:
... when opening the New-Dialog again.


DavidMStraub:
Makes sense, thanks for reporting & fixing!

PR #919, opened 2026-01-18
Edit repo ref

jittymolmathew92:


DavidMStraub:
Hi, I think a rebase is needed - the file `GrampsjsForMorePersonDetails.js` doesn't belong to this PR.

src/components/GrampsjsFormRepoRef.js
```diff
@@ -14,7 +14,27 @@ import './GrampsjsFormString.js'
 import {GrampsjsObjectForm} from './GrampsjsObjectForm.js'
 
 class GrampsjsFormRepoRef extends GrampsjsObjectForm {
+  static get properties() {
+    return {
+      repoRefKey: {type: String},
+      repoData: {type: Array},
```
DavidMStraub: Are you sure this is an array and not an object? (I'm not sure but I suspect it's an object.)

src/components/GrampsjsFormRepoRef.js
```diff
@@ -14,7 +14,27 @@ import './GrampsjsFormString.js'
...
+    }
+  }
+
+  constructor() {
+    super()
+    this.repoRefKey = ''
+    this.repoData = []
```
DavidMStraub: Same here.

DavidMStraub:
Hi,
the extension of `_rednerActionBtns` by `rowIndex` is suboptimal because my change in #914 with the new `identifier` was done exactly to use it for handle *or* index. Having a separate index argument is misleading.

If I understand correctly, you did it because you need the handle for `repoRefKey`. But I think the entire `repoRefKey` is redundant in `GrampsjsFormRepoRef`, because the handle of the repository that is references in repository reference is available as in `this.data.ref` in `GrampsjsFormRepoRef`.

So, my suggestion:

- revert the change to `_rednerActionBtns`
- remove the `repoRefKey` property and use `this.data.ref` instead

jittymolmathew92:
Updated and verified all the repo ref functionalities according to the identifier change in #914 and pushed

jittymolmathew92:
Added new PR for the same functionality

PR #925, opened 2026-01-25
Allow editing repository reference

DavidMStraub:


DavidMStraub:
I think all 4 issues identified by Copilot are sound. @jittymolmathew92 I will fix them now don't worry, but please have a look afterwards when you have time, I think it's useful.

PR #937, opened 2026-01-30
Sorting and adding more missing Bulgarian translations

iskrenpp:


DavidMStraub:
Thanks for this, but translations can only be updated via Weblate. Editing the JSONs directly breaks the sync.

https://www.grampsweb.org/development/frontend/translation/

Issue #891, opened 2025-12-29
[Bug] Select fields sometimes not wide enough even if there is enough space


V-donck:
### Bug Description

When editing info for reports, the input picklist fields are not always wide enough, even if there is enough space. Th image below is one of the many examples.
<img width="2540" height="1142" alt="Image" src="https://github.com/user-attachments/assets/bb144435-29d5-4872-b1f6-63b07de0c259" />

### Gramps Web Version Information

Gramps 6.0.6
Gramps Web API 3.6.0
Gramps Web Frontend 25.12.0
Gramps QL 0.4.0
Sifts 1.1.1
locale: en
multi-tree: false
task queue: true
OCR: true
chat: false

### Verification

- [x] I have searched for existing issues before creating this one

DavidMStraub:
We could increase the default width (as long as we ensure it doesn't exceed the viewport width on mobile).

Mid-term, the `mwc-select`s should be replaced by `md-filled-select`s (not sure if that changes anything, but what I mean is that it's not worth investing much work in tweaking the display before refactoring).

PR #959, opened 2026-02-26
Select fields sometimes not wide enough even if there is enough space

odiouschris:
Increase mdc-menu-min-width to 600px.
ref: #891

### DIFF
--- a/src/components/GrampsjsFormSelectObject.js
+++ b/src/components/GrampsjsFormSelectObject.js
@@ -35,7 +35,9 @@ class GrampsjsFormSelectObject extends GrampsjsAppStateMixin(LitElement) {
       css`
         mwc-menu {
           --mdc-menu-min-width: 200px;
-          --mdc-menu-max-width: 400px;
+          // .cca: Increase width of Source selection dropdown (increase to width of other dropdowns accepted.)
+          --mdc-menu-min-width: 600px;
+          //--mdc-menu-max-width: 400px;
         }
 
         .container {


src/components/GrampsjsFormSelectObject.js
```diff
@@ -35,7 +35,9 @@ class GrampsjsFormSelectObject extends GrampsjsAppStateMixin(LitElement) {
       css`
         mwc-menu {
           --mdc-menu-min-width: 200px;
-          --mdc-menu-max-width: 400px;
+          // .cca: Increase width of Source selection dropdown (increase to width of other dropdowns accepted.)
+          --mdc-menu-min-width: 600px;
```
DavidMStraub, line 39: Line 39 makes line 37 obsolete

src/components/GrampsjsFormSelectObject.js
```diff
@@ -35,7 +35,9 @@ class GrampsjsFormSelectObject extends GrampsjsAppStateMixin(LitElement) {
       css`
         mwc-menu {
           --mdc-menu-min-width: 200px;
-          --mdc-menu-max-width: 400px;
+          // .cca: Increase width of Source selection dropdown (increase to width of other dropdowns accepted.)
```
DavidMStraub, line 38: Those comments are noise, the code is self-explanatory

src/components/GrampsjsFormSelectObject.js
```diff
@@ -35,7 +35,9 @@ class GrampsjsFormSelectObject extends GrampsjsAppStateMixin(LitElement) {
...
+          --mdc-menu-min-width: 600px;
+          //--mdc-menu-max-width: 400px;
```
DavidMStraub, line 40: No point in having a commented-out property here. Keep it or remove it. If you remove it, please check all (!!) dropdowns in Gramps Web to make sure they are don't become excessively wide (even on mobile!)

odiouschris: I had checked the width of this change on mobile devices, however, as stated in #891, the existing code already causes an overflow in the container:

image

Increasing the width by 200px (to 600px) improves usability until I understand how to update the to "respect" the parent dialog.

For my first code commit, you have been very instructive. Thank you.

PR #960, opened 2026-02-26
Fix missing replace media button for non image media (#951)

### LINKED ISSUE #951: [Bug] Cannot replace media (missing button)
### Bug Description

On document editor page if media type is application/pdf button for media replacement is missing:

Image

For example image/jpeg:

Image

Gramps 6.0.6
Gramps Web API 3.7.1
Gramps Web Frontend 26.1.0
Gramps QL 0.4.0
Sifts 1.1.1
locale: en
multi-tree: false
task queue: false
OCR: true
chat: false

### Verification

5m4u66y:
When editing a media object, the `Replace file" button was available only for images. This fixes #951 and allows to replace any file type.

### DIFF
--- a/src/components/GrampsjsMediaObject.js
+++ b/src/components/GrampsjsMediaObject.js
@@ -239,6 +239,12 @@ export class GrampsjsMediaObject extends GrampsjsObject {
             )}
       </grampsjs-rect-container>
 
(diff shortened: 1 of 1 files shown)


DavidMStraub:
Perfect, thanks!

PR #962, opened 2026-02-26
Fix get descendants to correctly filter birth children for maternal lines

5m4u66y:
Fix an issue in `getDescendantTree` where birth children were only filtered using `frel` (father relation).

This caused descendants to be missing when the current person was the mother in a family and the descendant relation to the father was not 'Birth' (ie. Stepchild). Maternal birth relationships are stored in `mrel` (mother relation).


DavidMStraub:
Thanks for the fix!

Issue #297, opened 2023-10-28
Show association references


DavidMStraub:
Currently, we only show the associations of a person, but we don't show the associations that *other* people have pointing to this person. A (potential) user gave feedback that this would be nice to have, because often e.g. godparent associations are not created bidirectionally but only on one of the two involved people.

vamsi-1111:
Hey! My name is Vamsi and I am a part of Codepath's AI103 program where my main goal is to meaningfully contribute to open source projects. This issue caught my eye and I would love to work on a fix!

DavidMStraub:
Alright thanks, let me know if you have questions.

Issue #29, opened 2021-09-30
Event Reference Information attributes are not displayed


brucejackson:
I am using the Forms Gramplet to capture Census Data.  The Gramplet creates a single event that is shared among all household members.  However, details about a specific household member are stored in the Event Reference Attributes (see image).  The attributes are not visible in gramps.js.  I believe these attributes are stored in the person data and not the event data.  

I would be a nice enhancement is these attributes could be seen in gramps.js.  

Thanks,
Bruce

![image](https://user-images.githubusercontent.com/1219029/135477460-eae1653c-8e39-4e12-b176-a7efe43a28c2.png)

DavidMStraub:
Yes, you're right; this applies to all the "secondary objects" in Gramps that can have notes and attributes: media references, event references, child references, repository references ... in Gramps' [data model](https://www.gramps-project.org/wiki/index.php/Gramps_Data_Model), there are even crazier things like notes for attributes. While it's great to have that flexibility, I find it very hard to come with an intuitive UI to display (let alone edit) that information. Any ideas?

cdhorn:
I tried to take a pass at doing that with the view I was playing around and experimenting with. You can find it at https://github.com/cdhorn/profileview  

It is misnamed, should probably call it linked mode since the idea was to be able to navigate across all the objects to be able to view and edit anything.

Secondary objects that contain other secondary objects like the event reference above kind of throw a monkey wrench in the navigation. I kind of set it aside when I hit that to take a break for a while to spend some time on other interests. Hope to get back to it in another month or so.

brucejackson:
This is my first thought about an approach (without yet looking at the ProfileView gramplet).  I'm going to use the same data as above for simplicity. 

Starting at the Event Tab for a person (https://host/person/I0394).  I have Increased the size of the icons so they can be activate links.  I changed how the table is linked.  For each row the columns Date, Type, Description, and Place all link to the Event.   The icons (photos, notes, etc) link directly to a Tab within the Events page (e.g. the photo icon leads directly to the Gallery tab in the event).  I have added an icon, details (![outline_details_black_24dp](https://user-images.githubusercontent.com/1219029/135722691-ee56a9b2-2aad-42a5-ba84-e820ff0336e6.png)).  This icon links to the EventRef information for this person.   

- Link to Event (Green):  https://host/events/E2129
- Link to Gallery (Red):  https://host/events/E2129?ActiveTab=Gallery
[...]

PR #976, opened 2026-03-05
Display source citations from associated events along with direct citations

5m4u66y:
### Desired feature

Currently the Sources tab of an object only displays direct citations attached to the object itself (citation_list).

However, many sources are attached to events associated with the object (e.g. birth, death, marriage). In those cases, the Sources tab appears empty even though the object has citations through its events.

This makes it difficult for users to quickly see and access all sources relevant to an object.

### Proposed Solution

This change makes the Sources tab also display citations coming from associated events, in addition to direct citations.

The behavior differs slightly between view and edit modes:
- In View mode displays both direct citations and citations attached to associated events
- In Edit mode displays only direct citations, since event citations cannot be edited from this view

### Implementation

1. Show the Sources tab when event citations exist
The tab visibility condition now also checks whether any associated event contains citations:
```javascript
(data?.extended?.events || []).some(
  event => (event.citation_list || []).length > 0
)
```
2. Get citations list
Direct and events citations are both available from the data object. But for events citation, we only have the handle instead of the gramps_id.

Direct citations are retrieved from:
```javascript
data.extended.citations
```

Event citations are retrieved from:
```javascript
data.extended.events[].citation_list
```

In view mode the two lists are merged and deduplicated using a Set.

3. Fetch citations using gql
Previously citations were fetched using the HasIdOf filter rule, which requires gramps_id.

Since event citations only provide handles, the request now uses the gql query parameter to filter citations by handle:
```javascript
handle="H1" or handle="H2" ...
```

### Result

The Sources tab now displays all citations associated with an object, including those attached to events. This implementation does not allow to distinguish if citations are directly attached to the object or coming from events. This is fine for my needs, but tell me if I need to find a way to improve this.

### DIFF
--- a/src/components/GrampsjsObject.js
+++ b/src/components/GrampsjsObject.js
@@ -119,7 +119,11 @@ const _allTabs = {
   },
   sourceCitations: {
     title: 'Sources',
-    condition: data => data?.citation_list?.length > 0,
+    condition: data =>
+      data?.citation_list?.length > 0 ||
+      (data?.extended?.events || []).some(
+        event => (event.citation_list || []).length > 0
+      ),
     conditionEdit: data => 'citation_list' in data,
   },
   citations: {
@@ -738,15 +742,28 @@ export class GrampsjsObject extends GrampsjsAppStateMixin(LitElement) {
           .profile=${this.data?.profile?.children}
           ?edit="${this.edit}"
         ></grampsjs-children>`
-      case 'sourceCitations':
+      case 'sourceCitations': {
+        const directCitations = (this.data?.extended?.citations || [])
+          .map(obj => obj.handle)
+          .filter(obj => Boolean(obj))
+        let citationIds = directCitations
+
+        // In edit mode, only show direct citations (which are editable)
+        // In view mode, combine with event citations
+        if (!this.edit) {
+          const eventCitations = (this.data?.extended?.events || [])
+            .flatMap(obj => obj.citation_list || [])
+            .filter(obj => Boolean(obj))
+          citationIds = [...new Set([...directCitations, ...eventCitations])]
+        }
+
         return html`<grampsjs-view-source-citations
           active
           .appState="${this.appState}"
           ?edit="${this.edit}"
-          .grampsIds=${(this.data?.extended?.citations || [])
-            .map(obj => obj.gramps_id)
-            .filter(obj => Boolean(obj))}
+          .grampsIds=${citationIds}
         ></grampsjs-view-source-citations>`
+      }
       case 'notes':
         return html` <grampsjs-view-object-notes
           active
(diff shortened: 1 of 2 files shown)


DavidMStraub:
Thanks for this, but please open an issue first; I don't want to review the code before discussing a solution.

Please also take into account that Gramps Web currently does not display *any* information about attributes, citation, or media references of secondary objects, see the related issues

https://github.com/gramps-project/gramps-web/issues/29
https://github.com/gramps-project/gramps-web/issues/297

Of course, it would be good to display all that, but I've always found it difficult to come up with a UI that does not become overcomplicated.

And, most importantly: I don't want to start displaying stuff that cannot be edited. This would be super confusing for users.

At the moment, Gramps Web is at least self consistent in the sense that it displays all information that can be edited in the UI, and vice versa.

5m4u66y:
Hi David,

Thanks for the feedback, and sorry for opening the PR without creating an issue first. I initially implemented this change because I needed the functionality myself, and the code modifications turned out to be simpler than I expected. But I agree that this is something that should be discussed first.

I understand the concern about maintaining consistency in the UI, especially the principle that Gramps Web should only display information that can be edited from the interface. That makes sense from a usability perspective.

My motivation for this change is that, in practice, users often attach sources to events (e.g. birth, death, marriage), and when viewing an object it can therefore look like there are no sources at all, even though relevant citations exist on its events. From a user perspective this can sometimes feel misleading.

That said, I also understand the concern about showing information that cannot be edited from the same place.

Do you think an approach like one of the following could be acceptable?

- Display direct citations first, then a separate section such as “Citations referenced in events”.
- Visually differentiate event citations (for example with a label or icon) so it is clear they come from associated events.
- Potentially make the event link explicit so users can navigate to the event where the citation can actually be edited.

Of course, if this does not fit the current UI direction of Gramps Web, I completely understand. I’m happy to discuss alternative approaches or help think about a more general solution that could also address the related issues you mentioned. But as you said, there is no easy solution I could think of right now.

Thanks again for taking the time to review and explain the current design considerations.

DavidMStraub:
Actually, I realize I might have been talking/thinking about something completely different, namely citations of *secondary* objects (event references vs events, media references vs media, etc.); but your PR is about displaying citations of *primary* objects linked to the person. Sorry for the misunderstanding. From a data model perspective, the event reference citations are "closer" to the person because they are part of the person object, while the event citations are part of the event object.

What I would definitely not do is display the "second-degree" citation by default, because this makes it harder for users to understand where those are stored - this will become particularly relevant when we implement more fine grained permissions (e.g. https://github.com/gramps-project/gramps-web-api/discussions/766).

But what we could do is to have a toggle switch "show event sources" or something like that, which could be the same UI as a toggle switch for the event list to display family events (marriages etc.).

DavidMStraub:
I thought about this and I think it's not something we should do, not even optionally. We should clearly distinguish between person and event citations as they have a different genealogical meaning. If a citation *is* relevant for both the person and the event, the correct pattern is to share it between both, which is already possible.

Issue #541, opened 2024-12-08
Modernize web components


DavidMStraub:
See sub-issues.

PR #982, opened 2026-03-08
Migrate mwc-fab to md-fab

DavidMStraub:
Replace all uses of @material/mwc-fab with @material/web/fab/fab.js. Icons use grampsjs-icon in slot="icon" with MDI paths. Colors are overridden via CSS custom properties in :root to match the previous mwc-fab appearance (blue background, white icon).


DavidMStraub:
Related to #541

Issue #596, opened 2025-02-04
Create persons from Family trees


ngtrian:
It is great to create a person directly from Family Trees, because it is quite intuitive to choose someone and add a new person like children, husband and wife or parents.

DavidMStraub:
I agree.

What is your opinion, would it be enough to have this in one of the tree views (e.g. the ancestor tree)?

If you like, it would also be interesting to research a little bit how this is done in other systems (including Gramps desktop).

jo-gross:
I'm also looking for a feature like that. For me the goal would be to simply give my grandma access and let her easily add more people herself. 
My first thought looks something like this:

<img width="612" alt="Image" src="https://github.com/user-attachments/assets/3f9c8a2f-9fad-46d6-906d-7625e9d315d8" />

Giving each card the option to add a relation/person. 
When clicking the plus, you could get asked about the relation like "mother, father, brother, sister, child, or what else" and the person like "search for existing one or create new". I´m not sure if this is the best way, but it´d improve the usability a lot.

monesje:
I would love this feature. I'm finding it really unintuitive to have to go back and manually add a new person, then navigate to the family and add the person to that family, just so they appear in the tree. So many manual steps!

DavidMStraub:
Please refrain from adding "yes please" comments. Use 👍 reactions if you agree and comment if you have something to contribute. Otherwise you just create noise and distract everybody involved, which will not increase the likelihood of this being implemented. I know the intentions are good, but it's frankly annoying.

Sebolains:
Just here to add my 2 cents, as someone who's somewhat against this feature. If this is implemented, I think it should be under some sort of "edit" mode for the charts, which have clearly been read-only until now. Or perhaps in a new type of chart meant for editing. Adding more elements to the default charts will (a) clutter the more common use of viewing the tree, and (b) interfere with a potential future export feature (#346).

cookie-monster1649:
I agree with that implementation @Sebolains. As someone who wants to keep the set up simple though, this is an absolutely essential feature that is missing from web. Without it, quickly and easily building trees just isn't possible without using the full Gramps app and syncing back into gramps-web. How exactly the mechanism works to access it, I don't mind. I just know it would be handy way beyond any kind of DNA stuff to me.

DavidMStraub:
The feature request is open, so there is no point in arguing about who needs it or not. Of course, the UI will have to be minimal enough to not cutter the tree, but that's a solvable problem.

[...]

PR #995, opened 2026-03-21
Feature/create person from family tree

jittymolmathew92:


DavidMStraub:
#596

DavidMStraub:
Thanks! Can you describe what the current status is? If I understand correctly, the UI is done but the API calls are incomplete (relationship not handled) and the edit events are not handled because we don't inherit from `GrampsjsViewObject.js`. If my understanding is correct, and can make some recommendations for next steps.

In the meantime I'll add some comments on the code.

src/views/GrampsjsViewTreeChart.js
```diff
@@ -1,13 +1,20 @@
...
 import '../components/GrampsjsTreeChart.js'
+import '../components/GrampsjsFormNewPerson.js'
 
 export class GrampsjsViewTreeChart extends GrampsjsViewTreeChartBase {
   constructor() {
     super()
     this._setAnc = true
     this.defaults.nAnc = 3
+    this._addPersonRelationDialogOpen = false
+    this._addPersonRelationDialogData = null
```
DavidMStraub: If you add properties in the constructor, they must first be declared in `properties` to become reactive!

DavidMStraub: Also, please don't use `null` for initial value, but use the correct type - `{}` if it's an object, `[]` if it's an array.

src/views/GrampsjsViewTreeChart.js
```diff
@@ -34,18 +41,121 @@ export class GrampsjsViewTreeChart extends GrampsjsViewTreeChartBase {
     this.nameDisplayFormat = this.defaults.nameDisplayFormat
   }
 
-  renderChart() {
+  _openAddPersonRelationDialog(data) {
+    this._addPersonRelationDialogOpen = true
+    this._addPersonRelationDialogData = data
+    this.requestUpdate()
```
DavidMStraub: Never call `requestUpdate` manually. I think this is only needed because you forgot to declare the properties (see above).

DavidMStraub: By the way, after removing `requestUpdate` and declaring the properties, many of the new handlers become 1-line and don't need to be separate methods - just set the property directly at the call site.

src/views/GrampsjsViewTreeChart.js
```diff
@@ -34,18 +41,121 @@ export class GrampsjsViewTreeChart extends GrampsjsViewTreeChartBase {
...
+
+    if (
+      !relatedPerson?.data ||
+      !['father', 'mother', 'child'].includes(this.relationship)
+    ) {
+      this._closeNewPersonFormDialog()
+      return
+    }
+
+    const createPerson = await this.appState.apiPost(
```
DavidMStraub: Please add a TODO comment here that the relation still needs to be handled - this will help our AI colleagues :)

src/views/GrampsjsViewTreeChart.js
```diff
@@ -34,18 +41,121 @@ export class GrampsjsViewTreeChart extends GrampsjsViewTreeChartBase {
...
+        >
+        </grampsjs-tree-chart>
+      </div>
+    `
+  }
+
+  render() {
+    return html`
+      ${this.renderChart()}
+      <md-dialog
```
DavidMStraub: Please extract this into a separate `renderAddPersonRelationDialog` for clarity

src/views/GrampsjsViewTreeChart.js
```diff
@@ -34,18 +41,121 @@ export class GrampsjsViewTreeChart extends GrampsjsViewTreeChartBase {
...
+          </md-text-button>
+          <md-text-button @click="${() => this._handleAddPerson('mother')}">
+            Add Mother
+          </md-text-button>
+          <md-text-button @click="${() => this._handleAddPerson('child')}">
+            Add Child
+          </md-text-button>
+        </div>
+      </md-dialog>
+      ${this._handleAddNewPersonPopup()}
```
DavidMStraub: this should not be called `_handle...` but `_render...` - it's not a handler

DavidMStraub:
The + signs should only be displayed if the user has role editor or above.

This can be achieved by passing a `canEdit` property to the view tree chart components, as done for other views.

One more thing: in my local testing, the button par on the top left of the tree chart (select number of generations to show etc.) was not visible when I tested your branch. Not sure why though.

src/views/GrampsjsViewTreeChart.js
```diff
@@ -34,18 +41,121 @@ export class GrampsjsViewTreeChart extends GrampsjsViewTreeChartBase {
...
+  render() {
+    return html`
+      ${this.renderChart()}
+      <md-dialog
+        ?open="${this._addPersonRelationDialogOpen}"
+        @close="${() => this._closeAddPersonRelationDialog()}"
       >
-      </grampsjs-tree-chart>
+        <div slot="headline">${this._('Add Person with Relation')}</div>
+        <div slot="actions">
```
DavidMStraub: I'd just makes this an `md-list` with `md-list-item`s of type `"button"` - this will make it easier to add more in the future.

jittymolmathew92:
> Thanks! Can you describe what the current status is? If I understand correctly, the UI is done but the API calls are incomplete (relationship not handled) and the edit events are not handled because we don't inherit from `GrampsjsViewObject.js`. If my understanding is correct, and can make some recommendations for next steps.
> 
> In the meantime I'll add some comments on the code.

jittymolmathew92:
Accidently closed

jittymolmathew92:
I did the review comments and pushed the code and it just the UI only. I think after deciding the API how to manage relationship and we can further proceed with it.

src/components/GrampsjsTreeChart.js
```diff
@@ -98,6 +100,7 @@ class GrampsjsTreeChart extends GrampsjsChartBase {
         bboxWidth: this.containerWidth,
         bboxHeight: this.containerHeight,
         nameDisplayFormat: this.nameDisplayFormat,
+        canEdit: this.canEdit,
```
DavidMStraub: I realized we don't need `canEdit` as a property of `GrampsjsTreeChart` since it's contained in `appState`: this line can just be 
`canEdit: this.appState.permissions.canEdit`

DavidMStraub:
Superseded by #1040

PR #991, opened 2026-03-14
Replace mwc-dialog with md-dialog everywhere

DavidMStraub:
[AI generated summary]

This pull request migrates all dialogs in the codebase from the deprecated `@material/mwc-dialog` to the new `@material/web/dialog` (`md-dialog`). This update includes refactoring dialog usage across multiple components, updating dialog styling, and removing the old dependency. The changes modernize the UI, improve accessibility, and ensure compatibility with the latest Material Web components.

**Dialog Migration and Refactoring:**

[...]

PR #997, opened 2026-03-23
Fix select dropdown clipped by dialog overflow

zlant:
Override overflow on md-dialog's internal .container and .scroller elements to prevent them from clipping the md-filled-select dropdown menu. Also set menu-positioning="fixed" on the select component.

Before:
image

After:
image

### DIFF
--- a/src/components/GrampsjsFormSelectType.js
+++ b/src/components/GrampsjsFormSelectType.js
@@ -127,6 +127,7 @@ class GrampsjsFormSelectType extends GrampsjsAppStateMixin(LitElement) {
               label="${this.label}"
               .value="${this.value}"
               id="select-type"
+              menu-positioning="fixed"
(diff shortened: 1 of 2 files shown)


DavidMStraub:
Interesting hack :slightly_smiling_face: 

If this works reliably, we should think about subclassing `md-dialog` instead.

I am about to merge this #991 which will use the new component *everywhere*.

DavidMStraub:
Did you check whether `menu-positioning="fixed"` is enough to achieve the desired effect?

DavidMStraub:
Actually, I don't see this clipping happening at all - neither on Firefox nor on Chrome:

grafik

zlant:
This issue appears in Safari 16.5. Sorry I didn't mention it earlier. You can close this PR.

PR #1001, opened 2026-03-24
feat: add user-default family tree view and shared display preferences

elig-45:
## Problem
The Family Tree entry point does not let users choose which tree tab opens by default. Users who prefer a different tab (for example Fan or Relationship) must switch every time.

## Summary of changes
- Added a **Family tree preferences** section in **Settings > User settings**.
- Added one user-level setting: `treeDefaultView`.
- Added a single control: **Default family tree view**.
- Applied the preferred tab when entering the Family Tree page.
- Added unit tests for tree view tab index mapping/fallback.

## Not included (intentional)
- No duplicate chart controls (max depth, max images, name display format).
- Those are already persisted by #911 and remain unchanged.

## Linked issue
- Related to #67

## Screenshots
Screenshots are attached in PR discussion comments.

## Testing
- `npx eslint src/treeDefaults.js src/views/GrampsjsViewSettingsUser.js src/views/GrampsjsViewTree.js test/unit/treeDefaults.test.js`
- `npx prettier src/treeDefaults.js src/views/GrampsjsViewSettingsUser.js src/views/GrampsjsViewTree.js test/unit/treeDefaults.test.js lang/en.json --check`
- `npx vitest run test/unit/treeDefaults.test.js`
- `npm run build`

### DIFF
--- a/src/views/GrampsjsViewTree.js
+++ b/src/views/GrampsjsViewTree.js
@@ -58,6 +58,7 @@ export class GrampsjsViewTree extends GrampsjsView {
     this.view = 'ancestor'
     this._history = this.grampsId ? [this.grampsId] : []
     this._currentTabId = getTreeViewTabIndex(DEFAULT_TREE_VIEW)
+    this._appliedTreeDefaultView = null
   }
 
   renderContent() {
@@ -262,16 +263,18 @@ export class GrampsjsViewTree extends GrampsjsView {
       // limit history to 100 people
       this._history = this._history.slice(-100)
     }
-    if (changed.has('active')) {
-      if (this.active) {
-        this._setPreferredTab()
-      }
+    if (this.active && (changed.has('active') || changed.has('settings'))) {
+      this._applyPreferredTabIfNeeded()
     }
   }
 
-  _setPreferredTab() {
+  _applyPreferredTabIfNeeded() {
     const preferredView = this.settings?.treeDefaultView ?? DEFAULT_TREE_VIEW
+    if (preferredView === this._appliedTreeDefaultView) {
+      return
+    }
     const preferredIndex = getTreeViewTabIndex(preferredView)
+    this._appliedTreeDefaultView = preferredView
     if (this._currentTabId !== preferredIndex) {
       this._currentTabId = preferredIndex
     }


DavidMStraub:
Thanks for this, but please add more detail to this PR so I understand what it's about without having to dig through the code. If it's a feature that does not have a corresponding feature request, please also add screenshots.

elig-45:
Thanks for the feedback. I cleaned up the PR so it now contains only the tree-preferences feature, expanded the PR description with implementation details, linked issue #67, and added screenshots.

DavidMStraub:
Thanks for the additional info!

In my opinion, this duplicates - and actually clashes with - #911, which implemented persistence of chart settings. #67 was already marked as closed. Why do you think it's necessary to add dedicated settings?

elig-45:
Yep you're right, that makes sense. #911 already covers persistence of per-chart controls (in browser local storage only).

I’ll narrow this PR to the missing part only: selecting/persisting the default tab when entering the Family Tree page (treeDefaultView), without duplicating chart setting controls from #911.

lang/en.json
```diff
@@ -197,6 +197,8 @@
   "Nobody": "Nobody",
   "User group": "User group",
   "User Information": "User Information",
+  "Family Tree preferences": "Family Tree preferences",
+  "Default Family Tree view": "Default Family Tree view",
```
DavidMStraub: Your capitalization in this strings is not consistent.

DavidMStraub:
Ok thanks!

Comments:

- the screenshots were meant for PR discussion, do not commit them to git please
- the logic for tab handling is complicated with `_tabChangedByUser` and all that. Why can't we just set the preferred tab in the constructor or in `firstUpdated` and get rid of all the update callback stuff?

lang/en.json
elig-45: Ok I am correcting this right now.

elig-45: Family Tree preferences -> Family tree preferences
Default Family Tree view -> Default family tree view



Do you prefer like that ? @DavidMStraub

elig-45:
Update on your comments :

- Removed screenshot files from git history (no screenshots are committed in this PR).
- Simplified tree tab handling by removing _tabChangedByUser and applying 	treeDefaultView only when entering the Family Tree page.
- Normalized capitalization of the new user settings labels to match existing style.

lang/en.json
elig-45: Capitalization updated to sentence case to match existing Settings labels (`Family tree preferences`, `Default family tree view`).

DavidMStraub: Looks good, I'd say either all or none capitalized (we're not internally consistent overall)

elig-45:
Screenshots :

**User settings: Family tree preferences** User settings - Family tree preferences
**Family Tree default tab applied (Fan Chart)** Tree view - Fan Chart selected

DavidMStraub:
With your new logic, if I understand it correctly, if I open the tree view, then navigate to a person, then go back, it will always switch to the "preferred view". I think this can be annoying. What I meant was: when the tree view is first opened, switch to the default; when opening it next time, do *not* change the tab. Is there a showstopper to achieve this behaviour?

elig-45:
> With your new logic, if I understand it correctly, if I open the tree view, then navigate to a person, then go back, it will always switch to the "preferred view". I think this can be annoying. What I meant was: when the tree view is first opened, switch to the default; when opening it next time, do _not_ change the tab. Is there a showstopper to achieve this behaviour?

OK for me, I can adapt to :

Use `treeDefaultView` only the first time the Family Tree page is opened after the app loads.
After that, keep the tab the user last selected when they leave and return to the Tree page (including navigation to a person and back).

So it won’t revert to the preferred tab every time.
Regarding changes to `treeDefaultView` in Settings during the same session:
I’ll make the new value apply immediately. Does that work for you?

elig-45:
Btw thanks for the amazing work of maintaining Gramps web it's a great tool ! @DavidMStraub

DavidMStraub:
> Use treeDefaultView only the first time the Family Tree page is opened after the app loads.
After that, keep the tab the user last selected when they leave and return to the Tree page (including navigation to a person and back).
> 
> So it won’t revert to the preferred tab every time.
Regarding changes to treeDefaultView in Settings during the same session:
I’ll make the new value apply immediately. Does that work for you?

Yes, sounds good!

If we can achieve both of them without too complicated update callback logic and additional properties, it would be awesome!

elig-45:
Implemented with simple logic and no new reactive properties:

- `treeDefaultView` is applied the first time Tree is opened after app load.
- After that, returning to Tree keeps the last user-selected tab (including person navigation/back).
- If `treeDefaultView` changes during the same session, the new value is applied immediately (when Tree is active) and on next Tree entry if changed while inactive.

Implementation detail: I use one internal non-reactive field to track the last applied default value, and a small `update()` condition on `active/settings` to avoid complex tab-state callbacks.

src/views/GrampsjsViewTree.js
```diff
@@ -261,6 +263,21 @@ export class GrampsjsViewTree extends GrampsjsView {
...
+    if (this.active && (changed.has('active') || changed.has('settings'))) {
+      this._applyPreferredTabIfNeeded()
+    }
+  }
+
+  _applyPreferredTabIfNeeded() {
+    const preferredView = this.settings?.treeDefaultView ?? DEFAULT_TREE_VIEW
+    if (preferredView === this._appliedTreeDefaultView) {
+      return
+    }
```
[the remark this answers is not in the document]
DavidMStraub: Nope, because that' what we're trying to avoid.

Issue #1011, opened 2026-03-29
[Bug]  Export and other async tasks fail to update status on UI


artisanalcode:
### Bug Description

When triggering an export from the UI (or any async task), the GET /api/tasks/{taskId} polling requests are sent without the Authorization: Bearer header, causing 401 errors. Seen on GrampsHub.

**Root Cause Analysis**
1) `<grampsjs-task-progress-indicator>` extends `GrampsjsAppStateMixin` and relies on `this.appState.auth` to authenticate API calls. 
2) However, no parent component passes .appState to the component. Thus, `this.appState` defaults to an empty object: `{}`. So,  `this.appState.auth` is undefined, and when `apiGet` in `src/api.js` calls `auth.getValidAccessToken()`
3) The error fails silently (an empty catch {} block)
4) The request is submitted with no Authorization header.

It seems most other components using `GrampsjsAppStateMixin` receive `appState` from its parent.

**Steps to reproduce**

1) Log in to a GrampsHub (or any authenticated Gramps Web instance)
2) Navigate to Export
3) Click Generate
4) Open browser dev tools, Network tab
5) Reload page
6) Component spinner continues to spin, never completing task
7) Observe the `GET` `/api/tasks/{taskId}` requests fail with no authorization header
8) You can test the `GET` `/api/tasks/{taskId}`  with postman, the API endpoint works well

### Gramps Web Version Information

Gramps 6.0.7
Gramps Web API 3.9.0
Gramps Web Frontend 26.3.0
Host: grampshub.com

### Verification

- [x] I have searched for existing issues before creating this one

artisanalcode:
Opened PR: https://github.com/gramps-project/gramps-web/pull/1012

DavidMStraub:
FYI there is also a backend issue affecting background tasks, about to fix it:
https://github.com/DavidMStraub/gramps-web-sync/issues/55

PR #1012, opened 2026-03-29
Fix: Pass appState to grampsjs-task-progress-indicator in all views

artisanalcode:
**Description**

When triggering an export from the UI (or any async task), the GET /api/tasks/{taskId} polling requests are sent without the Authorization: Bearer header, causing 401 errors. Seen on GrampsHub.

Added .appState="${this.appState}" to every <grampsjs-task-progress-indicator> instance, matching the pattern already used by all other child components that rely on GrampsjsAppStateMixin.

Note: Added a basic unit test. But the empty catch needs to be improved and better testing added for further reliability.

Closes [Issue 1011](https://github.com/gramps-project/gramps-web/issues/1011)

**Screenshots**
Screenshot 2026-03-29 at 17 05 42
Screenshot 2026-03-29 at 17 05 28

### DIFF
--- a/test/unit/api.test.js
+++ b/test/unit/api.test.js
@@ -1,0 +1,47 @@
+import {describe, it, expect, beforeEach, afterEach, vi} from 'vitest'
+import {apiGet} from '../../src/api.js'
+
+describe('apiGet authentication', () => {
+  beforeEach(() => {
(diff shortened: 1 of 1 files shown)


DavidMStraub:
D'oh, thanks for the quick diagnosis and fix! My bad...

PR #1040, opened 2026-04-04
Add Family member to Ancestor Tree

jittymolmathew92:
The UI implementation for add new family member against a person is included in this PR.

There is no backend integration.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -289,5 +289,8 @@
   "Select files": "Select files",
   "Failed to upload": "Failed to upload",
   "Failed to update": "Failed to update",
-  "Not private": "Not private"
+  "Not private": "Not private",
+  "Add Family Member": "Add Family Member",
+  "Son": "Son",
+  "Daughter": "Daughter"
 }
--- a/src/charts/TreeChart.js
+++ b/src/charts/TreeChart.js
@@ -176,13 +176,7 @@ function TreeChartCore(
     .attr('id', d => d.data.id) // Unique id for each slice
 
   function triangleClicked(e) {
-    chart.node().dispatchEvent(
-      new CustomEvent('pedigree:show-children', {
-        bubbles: true,
-        composed: true,
-        detail: {pageX: e.pageX, pageY: e.pageY},
-      })
-    )
+    fireEvent(this, 'pedigree:show-children', {pageX: e.pageX, pageY: e.pageY})
     e.stopPropagation()
     e.preventDefault()
   }
--- a/src/charts/TreeChart.js
+++ b/src/charts/TreeChart.js
@@ -54,6 +54,7 @@ function TreeChartCore(
     getImageUrl = null,
     orientation = 'LTR',
     nameDisplayFormat = chartNameDisplayFormat.surnameThenGiven,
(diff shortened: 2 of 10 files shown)


lang/en.json
```diff
@@ -281,5 +281,10 @@
   "Uploading file %s of %s": "Uploading file %s of %s",
   "Select files": "Select files",
   "Failed to upload": "Failed to upload",
-  "Failed to update": "Failed to update"
+  "Failed to update": "Failed to update",
+  "Add Family Member": "Add Family Member",
+  "Father": "Father",
```
DavidMStraub: Those strings exist in the backend. Check `strings.js`. (It's expained in `CLAUDE.md` as well as the docs.)

src/charts/TreeChart.js
```diff
@@ -297,6 +298,43 @@ function TreeChartCore(
     .attr('paint-order', 'stroke')
     .text(d => clipString(`†${d.data.person.profile.death.date}`, textWidth(d)))
 
+  function addPerson(event, nodeData) {
+    chart.node().dispatchEvent(
+      new CustomEvent('add-new-person-relation', {
```
DavidMStraub: You can use `fireEvent` from `util.js`

src/components/GrampsjsTreeChartAddPerson.js
```diff
@@ -0,0 +1,195 @@
...
+      fireEvent(this, 'grampsjs:error', {message: createPerson.error})
+      return
+    }
+
+    e.preventDefault()
+    e.stopPropagation()
+    this._newPersonFormDialogOpen = false
+    this._addPersonRelationDialogData = {}
+    this._relationship = ''
+    fireEvent(this, 'db:changed')
```
DavidMStraub: This event is already fired by the `apiPost` call so this is a duplication

DavidMStraub:
Some UX considerations:

- the view needs an "edit mode" (with pencil FAB, blue app bar etc., like the object views) - plus signs should only displayed in tedit mode
- We should not offer "father" and "mother" as options to people who already have father and mother. While people can have multiple parent families (when adopted etc.), this is not something that is unique enough to handle from a tree view.

Next, you can implement the backend logic. But this is too complicated to implement from scratch: it can all be done via firing `edit:action` events. The problem is that these events are currently handled in `GrampjsViewObject`, from which `GrampsjsViewTreeChartBase` does not inherit. So it's best to move this handling logic somewhere up in the stack (perhaps even `GrampsJs.js`) or into a mixing. I suggest to feed this comment as prompt to Claude and ask it to do a tradeoff analysis of possible architectures and recommend one.

src/components/GrampsjsTreeChartAddPerson.js
jittymolmathew92: Removed unwanted fireEvent

src/charts/TreeChart.js
jittymolmathew92: Done

lang/en.json
jittymolmathew92: Done

DavidMStraub:
Superseded by #1111.

Issue #802, opened 2025-10-27
Person View: Distinguish person from siblings


reuter-max:
**Is your feature request related to a problem? Please describe.**
I haven't found an open Issue regarding this, so I thought I just report it myself 😃
When you create a new person it's shown as a sibling of itself, which might be correct in terms of data-relation, but it's imho rather confusing. One might think the person has a brother/sister with the same name.
Is it possible to hide the siblings-section within the person-view when the person has no siblings at all? Or maybe hide the person from the list of sibling, when the person do has sibling? 

**Describe the solution you'd like**
As mentioned above: either hide the person from its own list of siblings if present or maybe hide the whole siblings-section if no (other) siblings are present

**Describe alternatives you've considered**
Can't think of any other alternatives or existing settings which might affect this behaviour.

**Additional context**
_Example + Screenshot:_
My grandfather has 3 siblings, but within the section there are 4 entries with one pointing to himself. 

<img width="1917" height="945" alt="Image" src="https://github.com/user-attachments/assets/dba0b387-37b1-42fe-93d5-64a44af37594" />

_System Information_
Gramps 6.0.4
Gramps Web API 3.3.0
Gramps Web Frontend 25.8.0
Gramps QL 0.4.0
Sifts 1.0.0
locale: en
multi-tree: false
task queue: false
OCR: true
chat: false

DavidMStraub:
Hi,

you're right the current situation is suboptimal.

Actually, the sibling view is very similar to Gramps Desktop:

<img width="821" height="302" alt="Image" src="https://github.com/user-attachments/assets/688fbc72-cd46-4320-b900-ed03c47fc348" />

But there, the active person is highlighted in boldface. I think some kind of highlight should be enough to make this clearer. I do want to see the person in the list of their parents' children because it's important to know whether they were the firstborn, lastborn, etc.

Technical comment: the current component is a non-responsive table and it's quite likely this will reimplemented in the near future, so I don't think it's worth investing work in styling before that.

But the feature request is legit and can stay open until then.

DavidMStraub:
Changed the title of the feature request as we won't hide the person.

kinslayer1982:
> But there, the active person is highlighted in boldface. I think some kind of highlight should be enough to make this clearer. I do want to see the person in the list of their parents' children because it's important to know whether they were the firstborn, lastborn, etc.

But wouldn't it be enough to have the person displayed in the family view instead of having it displayed in the person's view, too?
[...]

PR #1069, opened 2026-04-11
Add a visual cue for a person's self in the sibling list.

efi:
Hi everyone, I am just learning how Gramps Web works on the dev side and also how to correctly contribute here...

This PR has some minimal changes that (finally) highlights the currently viewed person among its siblings in the list.

image

I hope the color I chose is considered tasteful and to reuse `grampsjs-color-skeleton-shine` is allowed. It should be always a bit lighter than the hover color.

Any hints for being a better contributor in the future is welcome of course.

Best regards!

### DIFF
--- a/src/components/GrampsjsChildren.js
+++ b/src/components/GrampsjsChildren.js
@@ -1,4 +1,4 @@
-import {html} from 'lit'
+import {html, css} from 'lit'
 
 import '@material/mwc-button'
 import '@material/mwc-icon-button'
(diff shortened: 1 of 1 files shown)


efi:
Seems quite related to issue #802 - also since the PR should make the number of "actual" siblings from that relationship more clear since you can visually count all the non-highlighted lines

DavidMStraub:
Thanks a lot for this!

Regarding your question about how to contribute best: in general it's always best to comment on existing issues before submitting a PR to avoid duplicate work. In this case, as I wrote in the linked issue

> Technical comment: the current component is a non-responsive table and it's quite likely this will reimplemented in the near future, so I don't think it's worth investing work in styling before that.

In fact, with the PRs I submitted in the last few days (like #1067) I was preparing exactly this reimplemntation, so I won't merge this improvement of a component that will be completely revamped for the next release. Sorry, but I'm sure you'll find something else to contribute soon :raised_hands:

efi:
Ah, that's completely fine, as long as the feature itself will be included, and sorry for the additional work now!

To be honest I only found the issue after filing the PR. :D - it was just something that always bugged me and it seemed to be a good easy fix to get started, so I got a bit ahead of myself...

PR #1071, opened 2026-04-11
Add SVG title element to cells in fan view to enable tooltips

efi:
Well, I hope this one does not interfere with any planned changes. I could not find issues regarding that topic but think it's a helpful addition:

image

### DIFF
--- a/src/charts/FanChart.js
+++ b/src/charts/FanChart.js
@@ -380,6 +380,14 @@ export function FanChart(
     .style('cursor', 'pointer')
     .on('click', clicked)
 
+  cell
+    .append('title')
+    .text(d =>
+      nameDisplayFormat === chartNameDisplayFormat.surnameThenGiven
+        ? '' + d.data.name_surname + ', ' + d.data.name_given
+        : '' + d.data.name_given + ' ' + d.data.name_surname
+    )
+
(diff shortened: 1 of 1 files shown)


DavidMStraub:
Good idea!

DavidMStraub:
One question, in which browsers have you tested it so far?

efi:
> One question, in which browsers have you tested it so far?

Chrome, Firefox and Edge on Windows.

It should really be quite universal since the feature is stable and supported for ages now: https://caniuse.com/mdn-svg_elements_title_tooltip_display

DavidMStraub:
Thanks!

I think copilot is right about the string fallback, otherwise LGTM!

efi:
Thank you!

Yes, I did not think that the names could be false / no string. However there are also other D3 operations adding text to the cell that don't handle these cases. Should I change those instances too, in this PR or should I not alter this one at all and fix everything in a new PR once this got through?

DavidMStraub:
Empty last names are pretty common - can you check what happens? Maybe it's handled by the caller, I don't remember...

efi:
> Empty last names are pretty common - can you check what happens? Maybe it's handled by the caller, I don't remember...

I could not produce anything but an empty string with any combination of missing / incomplete names for a person. So it seems to be **safe as it is now**. And it's a good thing anyways, that the logic for `name_surname` and `name_given` handles those cases instead of any d3 component.

Issue #1103, opened 2026-04-23
[Bug] Place hover shows HTML tag instead of only the place


MichaelNeys:
### Bug Description

When hovering over the place there is a Marker popup showing the name of the place but it shows the entire HTML tag as shown in the image. It should only show the name of the place but keep the link when you click on it.

<img width="348" height="110" alt="Image" src="https://github.com/user-attachments/assets/e9547687-64c1-47a9-9a79-c983f15d57ab" />

### Gramps Web Version Information

Gramps 6.0.4
Gramps Web API 3.12.0
Gramps Web Frontend [VI]{version}[/VI]
Gramps QL 0.4.0
Sifts 1.3.1
locale: en
multi-tree: false
task queue: true
OCR: true
chat: false

### Verification

- [x] I have searched for existing issues before creating this one

MichaelNeys:
Now it should only show the place name by using regex to strip out the place name out of the HTML tag. #1104 

<img width="348" height="110" alt="Image" src="https://github.com/user-attachments/assets/1ebc2704-332f-424d-8259-18ad9b5a4854" />

PR #1104, opened 2026-04-23
Strip out place name of hover popup on map for places

MichaelNeys:
Fixed place hover popup on map, the name is stripped out of the HTML tag using regex (fixes issue #1103 )

### DIFF
--- a/src/components/GrampsjsMapMarker.js
+++ b/src/components/GrampsjsMapMarker.js
@@ -56,7 +56,7 @@ class GrampsjsMapMarker extends LitElement {
     el.addEventListener('click', this.clickHandler.bind(this))
     // Add popup if needed
     if (this.popupLabel) {
-      el.title = this.popupLabel
+      el.title = this.popupLabel.replace(/<[^>]*>?/gm, '')
     }
     // Create the marker
     this._marker = new maplibregl.Marker({element: el})


DavidMStraub:
Hi, thanks for the bug report and the fix, but making the place name a link and then stripping the tag out again doesn't make much sense.

MichaelNeys:
Correct that does not make much sense. The `popupLabel` is populated like this:

```javascript
popupLabel="<a href='place/${obj.profile.gramps_id}'>${obj.profile.name}</a>"
```

There is no other element there with only the name so I continued with this quick fix without changing anything else.

How would you rather see this fixed?

DavidMStraub:
Ideally, making the link clickable again. I suspect it's a Leaflet-> Maplibre migration regression.

DavidMStraub:
Actually, I don't see this issue in Firefox.

MichaelNeys:
That's strange. I have this issue with the latest Gramps Web release on the latest version of Firefox and the latest version of Chrome. Both instances display it the same without any extensions enabled.

It is an issue I have seen for a long time on my self-hosted instance (which is also running the latest version). I've only now got time to take a look at it.

MichaelNeys:
For me the link is also clickable and I see this when I clicked on a place:

image

It goes to the possition and fills in the search bar.

DavidMStraub:
Ok, there was a misunderstanding. The popup (on click) is correct, the title (on hover) is wrong.

But there is a more serious issue, the place detail box doesn't open in the new version. I'll fix it.

PR #1125, opened 2026-05-03
Fix issues with service worker on app update

DavidMStraub:


DavidMStraub:
Hopefully this will also fix https://github.com/gramps-project/gramps-web/discussions/1114.

PR #1197, opened 2026-05-27
Improve French translations

nicolasfeyer:
## Summary
  - Fix typos (Editeur→Éditeur, inscirption→inscription, à propose→à propos, Metadata→Métadonnées, etc.)
  - Correct mistranslations (Given Name First→Prénom en premier, Search URL→URL de recherche, enclosing place→lieu englobant, Records→registres, etc.)
  - Unify terminology (tag→étiquette, Mail/E-mail→courriel, médium→multimédia, arbre familial→arbre généalogique, etc.)
  - Fix grammar/agreement and missing articles

  Note: `lang/fr.json` is normally managed via Weblate; flagging in case these should be routed through Weblate instead.


DavidMStraub:
Hi, thanks for trying to improve the translations, but Gramps (also Gramps Web) uses Weblate for translations, not Github. Please see here for instructions: https://gramps-project.org/wiki/index.php/Translating_Gramps_using_Weblate
Help is welcome.

PR #1213, opened 2026-06-01
Add family tree preferences translation in French

elig-45:
When I did my pull request #1001 I forgot to add the French translation.

So here it is ;)


elig-45:
I have just completed several translations directly on Weblate, including the one related to this PR.

If I understand correctly, you prefer translations to be handled through Weblate rather than by editing the files directly on GitHub. Therefore, you can close this PR if needed, as I have also translated this key on Weblate.

DavidMStraub:
Indeed, we can't edit those files in the repo, it breaks the weblate sync.

PR #1257, opened 2026-06-16
Allow pasting files into upload form

jerive:
## Summary
- add paste-to-upload support in the upload form component
- consume paste only when clipboard contains accepted files
- respect existing upload constraints (disabled, accept, multiple) for pasted files

## Why
- users can paste screenshots/images directly into upload flows instead of first saving files locally
- avoids breaking normal text paste behavior in other inputs/components

## Implementation details
- update paste handler in GrampsjsFormUpload to:
  - ignore paste events when component is not visible or disabled
  - extract file clipboard items and filter using accept rules
  - append or replace based on multiple
  - call preventDefault/stopImmediatePropagation only when files are consumed
  - trigger preview + formdata:changed like regular file selection
- add _acceptsFile(file) helper supporting:
  - extension tokens (e.g. .jpg)
  - explicit MIME types (e.g. image/png)
  - wildcard MIME categories (e.g. image/*)

## Testing
- checked diagnostics for modified component: no errors
- validated branch contains only src/components/GrampsjsFormUpload.js changes for this fix

## Notes
- no API changes
- no translation changes

### DIFF
--- a/src/components/GrampsjsFormUpload.js
+++ b/src/components/GrampsjsFormUpload.js
@@ -249,33 +249,65 @@ class GrampsjsFormUpload extends GrampsjsAppStateMixin(LitElement) {
   }
 
   _handlePaste(event) {
-    if (!this._isVisible) {
+    if (!this._isVisible || this.disabled) {
       return
     }
-    const {items} = event.clipboardData
-    if (!items) {
+    const {items} = event.clipboardData || {}
+    if (!items?.length) {
       return
     }
-    // prevent other forms down on the same page from also handling the paste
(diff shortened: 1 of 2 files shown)


DavidMStraub:
Thanks for your contribution but

1. This is already supported - file an issue if it doesn't work as expected
2. See CONTRIBUTING.md: "**Before implementing a new feature,** please open an issue to discuss your proposal. This helps avoid duplicate work and ensures alignment with project goals."

jerive:
2. thx, will do !

PR #1259, opened 2026-06-17
Validate dates on input

DavidMStraub:


DavidMStraub:
Note to self: merge when tested

PR #1265, opened 2026-06-18
fix: add error handling to local storage updates in api.js

MatrixNeoKozak:
The `updateSettings` function in `src/api.js` performs `JSON.parse` on data retrieved from `localStorage` without protection against malformed JSON, which can lead to application errors or prevent settings from being updated. This change wraps the update logic in a `try-catch` block to ensure robust handling of potential parsing errors, adhering to the project's defensive coding patterns already used in other storage management functions like `getTreeConfig` and `cleanOldDrafts`.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -80,13 +80,18 @@ export function getSettings() {
 export function updateSettings(settings, tree = false) {
   const key = tree ? 'grampsjs_settings_tree' : 'grampsjs_settings'
   const settingString = localStorage.getItem(key)
-  const parsedSettings = JSON.parse(settingString) || {}
-  const treeId = getTreeId() || 'unknown'
-  const existingSettings = tree ? parsedSettings?.[treeId] : parsedSettings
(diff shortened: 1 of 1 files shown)


DavidMStraub:
Thanks for your contribution, but I have doubts.

This JSON is written by the app, what case are you defending against? If "something" corrupts local storage, crashing the app is more honest than swallowing the error - even worse, your fix is not firing an error message, so the user will think their setting change was applied successfully.

Let's please first discuss what the issue is and then agree on a fix. (See CONTRIBUTING.md.)

MatrixNeoKozak:
You raise a fair point — since this JSON is only ever written by the app itself, a parse failure would indicate something is already wrong (corrupted storage, manual tampering), and silently swallowing it could mislead the user into thinking their settings were saved when they weren’t. I should have surfaced an error to the user rather than just catching silently, or not changed this at all. I’ll close this PR — happy to open an issue first if I find a concrete repro case, per the contributing guidelines.

PR #1268, opened 2026-06-21
Add configurable custom navigation links

### LINKED ISSUE #501: Enhancement - Put a Link to a User Guide / Manual in the top nav bar.

kjiwa:
## Add configurable custom navigation links

Closes #501.

### Summary

Adds a `navLinks` frontend config option for adding custom links to the main navigation bar, and a `hideTreeLink` option to hide the built-in Family Tree link. Together they let a deployer add links to same-origin tools (e.g. an embedded tree viewer behind the same reverse proxy) or external pages such as a user manual, and optionally replace the built-in Family Tree link with one of their own.

Both options are read from `window.grampsjsConfig`, consistent with the existing `hideDNALink` / `hideRegisterLink` / `loginRedirect` options.

### Config

```javascript
window.grampsjsConfig = {
  hideTreeLink: true,
  navLinks: [
    {title: 'Family Tree', url: '/topola', icon: '<mdi svg path>'},
    {title: 'User manual', url: 'https://example.org/manual', target: '_blank'},
  ],
}
```

Per-link keys: `title` and `url` (required), `icon` (optional SVG path, defaults to a generic link icon), `target` (optional, defaults to `_self`).

### Design notes

- **`target` defaults to `_self`.** This is the load-bearing decision. Navigation uses `pwa-helpers` `installRouter`, which intercepts every same-origin anchor click that has no `target`/`download`/`rel="external"` and routes it in-app. A same-origin custom link (the common case for a reverse-proxied tool) would therefore be swallowed by the router and never reach the real page. Rendering the anchor with a `target` makes the browser perform a real navigation. `md-list-item` forwards `target` (but not `rel`), so `target` is the clean lever; `_self` keeps the link in the same tab.
- **`icon` is a raw SVG path string**, matching how `<grampsjs-icon>` already works. A name→path registry was avoided because `@mdi/js` isn't tree-shakeable by dynamic name; the alternatives are bundling the full icon set or maintaining a curated subset, both worse than letting the deployer paste a path.
- **`title` is rendered verbatim**, not passed through `_()`. Deployer-provided strings aren't in the translation catalog, and `_()` strips a leading underscore.
- **Placement:** custom links are appended at the end of the primary navigation group (before the first divider) and are never marked `selected`, since they point outside the app's own routes.
- **`hideTreeLink` mirrors `hideDNALink` exactly** — it gates the menu item, the keyboard-shortcut help entry, and the `g c` shortcut handler.
- **Invalid entries** (missing/non-string `title` or `url`) are skipped with a `console.warn` rather than rendering a broken link or throwing during render.

### Backward compatibility

Inert by default: with neither option set, behavior is unchanged.

### Testing

- New unit tests for `normalizeNavLinks` (`test/unit/navLinks.test.js`).
- Full suite green (`npm test`), `npm run lint` clean, production build (`npm run build`) succeeds.

### Docs

Companion docs PR: gramps-project/gramps-web-docs#82.

### Open question

Naming — `navLinks` vs. `customNavLinks`. Happy to rename if preferred.

### DIFF
--- a/src/GrampsJs.js
+++ b/src/GrampsJs.js
@@ -423,8 +423,10 @@ export class GrampsJs extends LitElement {
                 <dd>${this._('Home')}</dd>
                 <dt><span>g</span> <span>b</span></dt>
                 <dd>${this._('Blog')}</dd>
-                <dt><span>g</span> <span>c</span></dt>
-                <dd>${this._('Family Tree')}</dd>
+                ${this.appState.frontendConfig.hideTreeLink
+                  ? ''
+                  : html`<dt><span>g</span> <span>c</span></dt>
+                      <dd>${this._('Family Tree')}</dd>`}
                 <dt><span>g</span> <span>t</span></dt>
                 <dd>${this._('Timeline')}</dd>
                 <dt><span>g</span> <span>m</span></dt>
@@ -1233,7 +1235,9 @@ export class GrampsJs extends LitElement {
       } else if (e.key === 'm') {
         fireEvent(this, 'nav', {path: 'map'})
       } else if (e.key === 'c') {
-        fireEvent(this, 'nav', {path: 'tree'})
+        if (!this.appState.frontendConfig.hideTreeLink) {
+          fireEvent(this, 'nav', {path: 'tree'})
+        }
       } else if (e.key === 'r') {
         fireEvent(this, 'nav', {path: 'recent'})
       } else if (e.key === 'f') {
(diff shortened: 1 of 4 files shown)


DavidMStraub:
Thanks for your contribution.

But, to respect my time, please open a feature request before submitting code. The description of your PR is completely different from the issue you linked. See CONTRIBUTING.md.

kjiwa:
Understood, and sorry for jumping ahead — I've opened #1271 to discuss the feature first. The "Closes #501" link was a mistake; this is a broader change than that issue. Happy to resubmit once there's agreement on the approach.

PR #1289, opened 2026-06-30
Allow editing parent relationship type of an existing child reference

### LINKED ISSUE #1286: Allow editing the parent relationship type (frel/mrel) of an existing child reference
**Is your feature request related to a problem? Please describe.**

When viewing a family, the *Children* list shows each child's relationship to the father and mother (e.g. "Father: Adopted / Mother: Birth"), as added in #1132. These relationship types (Birth, Adopted, Stepchild, Foster, …) can be set when **adding** a child — both the "add new child" and "link existing child" dialogs include *Relationship to Father / Mother* selectors.

However, once a child is part of the family there is **no way to edit** these relationship types from the web UI. If a child was added with the wrong type, or the relationship later needs to change (e.g. recording that a child was adopted while keeping the biological parent in a separate family), the only options today are to delete and re-add the child, or to edit the child reference through the API.

**Describe the solution you'd like**

Enable the standard edit (pencil) action on the children list. Selecting a child and clicking the pencil would open the child-reference form pre-filled with the current `frel`/`mrel`, exactly like editing an existing place reference in `GrampsjsPlaceRefs`. The `GrampsjsEditableList` base already provides the edit affordance (`hasEdit`, the pencil button, the `_handleEdit` hook) — it is used by place references, names, attributes, etc., but is simply not enabled for the children list.

**Describe alternatives you've considered**

- Deleting and re-adding the child with the correct relationship — loses ordering and is error-prone.
- Editing via the REST API — not accessible to non-technical users.

**Additional context**

I have a small implementation ready that mirrors the existing `GrampsjsPlaceRefs` edit flow and can open a PR.

sabal202:
Closes #1286.

## Summary

The *Children* list on a family page shows each child's relationship to the father and mother (e.g. "Father: Adopted / Mother: Birth", added in #1132), and these types can be set when **adding** a child. However, there was no way to **edit** them afterwards — the only options were to delete and re-add the child or to use the API.

This adds the standard edit (pencil) action to the children list. Selecting a child and clicking the pencil opens the child-reference form pre-filled with the current `frel`/`mrel`, mirroring how `GrampsjsPlaceRefs` edits an existing place reference.

## Changes

- **`GrampsjsChildren`**: enable `hasEdit`; `_handleEdit` opens the child-reference form pre-filled with the selected reference; on save it fires a new `updateChildRef` action (index-based), mirroring `GrampsjsPlaceRefs`.
- **`GrampsjsFormChildRef`**: make the form dual-mode — the person selector is shown only when adding (`new`); the father/mother relationship selectors are pre-populated from `data` when editing. (`child-frel`/`child-mrel` are already handled in `GrampsjsObjectForm`.)
- **`GrampsjsViewObject`**: handle the `updateChildRef` action via `updateObjectByIndex` on `child_ref_list`.
- **`lang/en.json`**: add the "Edit child reference" dialog title (other locales via Weblate).

## Testing

- `npm run lint`, `npm run typecheck`, and `npm run build` all pass.
- Deployed to a live instance and verified the edit (pencil) action opens the child-reference dialog pre-populated with the current father/mother relationship types. Saving reuses the same `updateObjectByIndex` → family `PUT` path already used for editing place references, names, etc. The add / link-existing flows are unchanged.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -319,6 +319,7 @@
   "Any text": "Any text",
   "A list of text values. Each entry in the list represents one line of text.": "A list of text values. Each entry in the list represents one line of text.",
   "Edit repository reference": "Edit repository reference",
+  "Edit child reference": "Edit child reference",
   "Unsaved changes restored": "Unsaved changes restored",
   "Uploading file %s of %s": "Uploading file %s of %s",
   "Select files": "Select files",
--- a/lang/en.json
+++ b/lang/en.json
@@ -339,6 +339,7 @@
   "Any text": "Any text",
   "A list of text values. Each entry in the list represents one line of text.": "A list of text values. Each entry in the list represents one line of text.",
   "Edit repository reference": "Edit repository reference",
+  "Edit child reference": "Edit child reference",
   "Unsaved changes restored": "Unsaved changes restored",
   "Uploading file %s of %s": "Uploading file %s of %s",
   "Select files": "Select files",
--- a/lang/en.json
+++ b/lang/en.json
@@ -339,7 +339,6 @@
   "Any text": "Any text",
   "A list of text values. Each entry in the list represents one line of text.": "A list of text values. Each entry in the list represents one line of text.",
   "Edit repository reference": "Edit repository reference",
-  "Edit child reference": "Edit child reference",
   "Unsaved changes restored": "Unsaved changes restored",
   "Uploading file %s of %s": "Uploading file %s of %s",
   "Select files": "Select files",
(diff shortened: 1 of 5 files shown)


lang/en.json
```diff
@@ -319,6 +319,7 @@
   "Any text": "Any text",
   "A list of text values. Each entry in the list represents one line of text.": "A list of text values. Each entry in the list represents one line of text.",
   "Edit repository reference": "Edit repository reference",
+  "Edit child reference": "Edit child reference",
```
DavidMStraub: There is an existing backend string "Edit the child reference" - no need to add a new frontend string here which needs to be translated again. Add to strings.js.

DavidMStraub:
Thanks for this! Please see copilot's comment and make sure you've tested the modified child ref editor in the existing call sites.

DavidMStraub:
Can you bring this to the finish line so I can include it in the next release? Thanks!

PR #1299, opened 2026-07-08
auto-split pasted coordinate pairs into latitude/longitude fields

petermgrund:
Pasting a decimal latitude, longitude pair (for example, copied from Google Maps or OSM) into either coordinate field now fills both fields, instead of dumping the whole string into one field. Applies to the pick-on-map coordinate dialog (new and existing places) and the inline fields on the new place form. In the dialog, the map also jumps to the pasted location. Non-coordinate text continues to be pasted normally. It does not parse DMS pairs since that would need to involve some coordinate conversion feature.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "gramps-js",
-  "version": "26.6.2",
+  "version": "26.7.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "gramps-js",
-      "version": "26.6.2",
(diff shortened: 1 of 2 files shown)


DavidMStraub:
Cannot reproduce any of this in Firefox 152.

Which browser, version, Gramps Web version, etc.?

Screencap would help. Thanks

DavidMStraub:
Sorry, I read this as a bug report when I made the comment.

DavidMStraub:
Was this closed accidentally?

PR #1314, opened 2026-07-27
Added Genealogie Online as search website

indykoning:
This is a site i personally had a lot of luck with doing research on ancestors


DavidMStraub:
Thank you for the suggestion.

Do you have more information about this service and its sources?

indykoning:
This is part of https://genealogie.coret.org/ where https://www.genealogieonline.nl/ is focussed on family trees. So this can mainly be used to find more ancestors and descendants for the person you enter.

The data is crowdsourced by other researchers, so each tree will have their own sources. But you can see an example here:
https://www.genealogieonline.nl/pasqual-bruss-family-tree/I527845.php#tab-bronnen

PR #1315, opened 2026-07-27
Show all available variables dynamically

indykoning:
I missed the middlename, birth year and death year in the list.
It is much easier to dynamically determine the available variables

### DIFF
--- a/src/components/GrampsjsFormExternalSearch.js
+++ b/src/components/GrampsjsFormExternalSearch.js
@@ -386,7 +386,11 @@ class GrampsjsFormExternalSearch extends GrampsjsObjectForm {
           ${this._(
             'Enter a search URL with template variables for person data. Available variables:'
           )}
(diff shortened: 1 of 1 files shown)


DavidMStraub:
Thank you for the pull request. Please open a feature request first next time.

I am skeptical about dynamically showing all variables. Can you paste the result of this template here?

indykoning:
Absolutely!
this will result in `{{name_given}}, {{name_surname}}, {{name_middle}}, {{place_name}}, {{birth_year}}, {{death_year}}` being output

DavidMStraub:
Oh, I see. The object is custom built exactly for this component, so it's perfectly safe to iterate over the keys. That's what I was worried about. Thanks!

indykoning:
Lacking the context of
> The object is custom built exactly for this component

I can absolutely understand being hesitant about looping over the data, luckily that's not the case in this case!
Glad to help in a project adding so much value to me

PR #1342, opened 2026-08-19
Feat: Blood relatives graph view under Family Tree tab

ConteDeStGermain:
New Blood Relatives tab on the Family Tree page. 

The graph is based on whoever you set as home person.

The three-filter chain collapses to one rule. HasCommonAncestorWith(homePerson) matches anyone whose ancestor set intersects yours, and Gramps makes each person their own ancestor — so that's exactly ancestors(me) ∪ descendants(ancestors(me)), children included. No saved filters, no filter-editor setup, no IsDescendantOfFilterMatch (which can't be expressed inline anyway since it needs a named filter).

GrampsjsViewBloodRelativesChart.js — extends the relationship-graph view, so it inherits the graphviz renderer, controls, and edit mode; only _getPersonRules changes GrampsjsViewTree.js, treeDefaults.js, GrampsjsViewSettingsUser.js:326 — tab appended last, so existing tab indices and saved default-view settings are untouched GrampsjsViewTreeChartBase.js — _setSpouses checkbox row, following the existing _setAnc/_setDesc/_setSep pattern Spouse checkbox lives in the ⚙ Preferences dialog (where every other chart option is), persists per user, and refetches on toggle. Spouses can't be done with an inline rule (IsSpouseOfFilterMatch needs a named filter), so it collects partner handles from the blood set's families and fetches them via ?handles=, chunked at 100 to stay under gunicorn's ~4 kB request-line cap. Their kids need no special handling — a blood relative's children are blood relatives, already in the set.

Verified: lint, typecheck, unit tests for the spouse-collection logic, and a full production build to resolve the import graph. One pre-existing unrelated failure in timeline.test.js (fails on a clean checkout too — confirmed by stashing).

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -414,5 +414,7 @@
   "Draw a selection": "Draw a selection",
   "New Blog Post": "New Blog Post",
   "Blog Post": "Blog Post",
-  "Failed to fetch the Blog tag": "Failed to fetch the Blog tag"
+  "Failed to fetch the Blog tag": "Failed to fetch the Blog tag",
+  "Blood Relatives": "Blood Relatives",
+  "Include spouses": "Include spouses"
 }
--- a/src/treeDefaults.js
+++ b/src/treeDefaults.js
@@ -4,6 +4,7 @@ export const TREE_VIEWS = [
   'hourglass',
   'relationship',
   'fan',
+  'blood',
 ]
 
 export const DEFAULT_TREE_VIEW = 'ancestor'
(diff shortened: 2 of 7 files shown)


DavidMStraub:
Hi,

thanks for this contribution, butt according to CONRTIBUTING.md you need to open a feature request first,

Thank you!

ConteDeStGermain:
Hey so I've created a feature request, what are the next steps?

PR #1351, opened 2026-08-22
Fix empty namespace when deleting notes

jerive:
## Summary
- `GrampsjsViewNotes` never set `_objectsName`, unlike every other `GrampsjsViewObjectsBase` subclass (People, Sources, Citations, Events, Families, MediaObjects, Places).
- As a result, deleting notes sent `{namespace: '', handles: [...]}` to `POST /api/objects/delete-by-handle/` instead of `{namespace: 'notes', ...}`.
- Sets `this._objectsName = 'notes'` in the constructor to match the existing pattern.

## Test plan
- [ ] Delete one or more notes from the Notes list view and confirm the request payload now includes `namespace: "notes"` and the notes are actually deleted.

### DIFF
--- a/src/views/GrampsjsViewNotes.js
+++ b/src/views/GrampsjsViewNotes.js
@@ -13,6 +13,7 @@ import '../components/GrampsjsFilterPrivate.js'
 export class GrampsjsViewNotes extends GrampsjsViewObjectsBase {
   constructor() {
     super()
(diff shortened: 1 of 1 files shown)


DavidMStraub:
Good catch! Apparently I never deleted a note because my notes are always good 😂

PR #1356, opened 2026-08-23
Preselect the current type when editing an internet address

### LINKED ISSUE #1355: [Bug] Internet metadata edit modals don't preserve Type
### Bug Description

When you open the editing modal for an item in the Internet section of the Metadata for a Person, the Type dropdown is not preserved. This appears to be due to the API for that section sending back type as a string rather than an object containing a property named `string`.

Gramps 6.0.8
Gramps Web API 3.20.1
Gramps Web Frontend 26.8.0
Gramps QL 0.4.0
Sifts 1.3.1
locale: en
multi-tree: false
task queue: true
OCR: true
chat: false

### Verification

karlvr:
Fixes #1355

The URL edit form read the type as `type.string`, but the URL type arrives from the API as a plain string, so the dropdown always fell back to the empty option.

We are binding `.data` here but it doesn't look like the select component uses it; is this intentional or would another issue to tidy that up be useful?

### DIFF
--- a/src/components/GrampsjsFormEditUrl.js
+++ b/src/components/GrampsjsFormEditUrl.js
@@ -22,7 +22,6 @@ class GrampsjsFormEditUrl extends GrampsjsObjectForm {
           ?loadingTypes=${this.loadingTypes}
           .types="${this.types}"
           .typesLocale="${this.typesLocale}"
-          .data="${this.data}"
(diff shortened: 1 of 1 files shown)


karlvr:
Removed `.data` binding as per Copilot feedback.

PR #1359, opened 2026-08-24
Fix phone number and ftp links in Internet list

### LINKED ISSUE #1354: [Bug] Phone numbers in Internet section get wrong link protocol
### Bug Description

A Phone entry in the Internet section for a person gets an `https://` link instead of a `tel://` link. Perhaps the frontend could observe the type is a Phone type and use the `tel://` protocol? For that matter, maybe `ftp://` needs special handling too.

Gramps 6.0.8
Gramps Web API 3.20.1
Gramps Web Frontend 26.8.0
Gramps QL 0.4.0
Sifts 1.3.1
locale: en
multi-tree: false
task queue: true
OCR: true
chat: false

### Verification

karlvr:
Fixes #1354. Continues the heuristic approach of `isValidEmail`, adding a test for phone numbers (hopefully broad enough but not too broad), and support for the FTP type. The approach for email, phone and FTP is different because obviously we can't distinguish FTP site addresses using regex.

### DIFF
--- a/src/components/GrampsjsUrls.js
+++ b/src/components/GrampsjsUrls.js
@@ -15,17 +15,49 @@ function isValidEmail(email) {
   return re.test(email)
 }
 
-function fixUrl(input) {
-  let url = input
-  try {
-    url = new URL(input)
-  } catch (error) {
-    if (isValidEmail(input)) {
-      return `mailto:${input}`
-    }
-    return `https://${input}`
+function isPhoneNumber(input) {
+  // Digits and usual separators
+  if (!/^\+?[\d\s()./-]+$/.test(input)) {
+    return false
   }
-  return url
+  // Exclude IPv4 addresses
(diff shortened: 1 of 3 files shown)


DavidMStraub:
Looks good, but can you please add a couple of unit tests for the new utility functions? Thanks!

karlvr:
I've separated the functions out and added some tests. I changed the function arguments now that it's outside the context of the component, which is I think clearer in JavaScript. I put all of the tests through `fixUrl` and was able to get 100% coverage of the functions in `urlUtil.js`.

I noted the feedback from Copilot above and I've added detection of `javascript:` URLs and a simple mitigation: `#`. Perhaps a mitigation is warranted in the backend?

DavidMStraub:
Thanks! I think the allowlist is the right strategy for the moment.

Issue #36, opened 2021-10-21
Allow editing addresses


DavidMStraub:
For me personally, this has low priority, but should of course be added for completeness.

Airick73:
I would be interested in assisting with the implementation of attributes for events. I have been reading through the files you have modified and added as a means to better understand how you are implementing this feature but if you have any suggestions to point me in the right direction I would be happy to help.

DavidMStraub:
> I would be interested in assisting with the implementation of attributes for events. I have been reading through the files you have modified and added as a means to better understand how you are implementing this feature but if you have any suggestions to point me in the right direction I would be happy to help.

Thanks! I have already added support for adding attributes for all object types in 50613e409fee09f02d684eac5cac53f1e8bfa4a1, have you checked whether it works for you?

Airick73:
I do see that there is a tab for attributes but there is no CRUD ability from the UI. Does the UI for this still need to be developed?
Person:
![image](https://user-images.githubusercontent.com/72322466/200336260-da3763a9-1f94-4c00-bf61-8b3e874e4f02.png)
Event:
![image](https://user-images.githubusercontent.com/72322466/200336418-3c5223bc-140a-488f-b23b-08b1811c9bfc.png)

DavidMStraub:
Oops that's a bug. Will fix soon.

DavidMStraub:
Should work with the last commit.

krishna-raj-sr:
Allow adding/Editing Addresses

karlvr:
I am planning to take a stab at this. I think there's enough reference UI etc to make a decent job of it for first looks.

karlvr:
I have something working for this... at the moment I've put the phone number as the secondary line, and the date on a second secondary line if there is one.

<img width="527" height="143" alt="Image" src="https://github.com/user-attachments/assets/7b44bbfe-f4b9-4278-ab64-695da7d50687" />

The editing modal is pretty simple. I can attached Citations and Notes like Associations have... but I know there's a bunch of discussion around these (#29) and it seems like editing is the easy part, but perhaps something to leave out for now as there's no display solution?

DavidMStraub:
Implemented in #1360

PR #1360, opened 2026-08-24
Address editing

karlvr:
Relates to #36. This is based on #1352, to avoid conflicts. 2 of the 5 commits are from that PR.

I'm hoping to use Gramps Web in place of Gramps for my family, and keeping addresses is one of things we do. I've chosen a really simple implementation, which includes a proposal to change the way addresses are rendered; making them render more like the other editable types. I am of course happy to make changes to suit your direction.

image

And with a date:
image

Editing:
image

Editing modal:
image
image


DavidMStraub:
Please rebase onto main after #1352 is merged, thanks!

PR #1375, opened 2026-09-04
Replace `mwc-select` with `md-select` so dropdowns escape dialog clipping

### LINKED ISSUE #1363: [Bug] Select fields are clipped by dialog
### What did you do?

1. Open an Event
2. Edit
3. Edit Date & Place
4. Select "Type"

### What happened then?

You can see in the screenshot that the select is clipped in Safari.

Image

### What did you expect to happen?

Unclipped select so I could choose any option.

### Version information

Gramps 6.0.8
Gramps Web API 3.21.0
Gramps Web Frontend 26.8.0
Gramps QL 0.5.0
Sifts 1.3.1
locale: en
multi-tree: false
task queue: true
OCR: true
chat: false

### Possible cause or solution (optional)

I understand from #891 that a move from `mw-select` to `md-filled-select` is being considered. I believe that will fix this too; I am happy to do that conversion, it looks pretty mechanical.

### Before submitting

- [x] I wrote this myself, or used an AI assistant and have read the whole text and agree with it

karlvr:
## Linked issue

Closes #1363

## What this changes

`mwc-select` renders its menu inside its own shadow root, so in Safari the menu is clipped by `md-dialog`'s `.container` (`overflow: hidden` plus `border-radius`) even with `fixedMenuPosition` set. `md-select` defaults to `menu-positioning="popover"`, which puts the menu in the top layer where no ancestor can clip it.

The PR converts all `mwc-select` instances to `md-filled-select`/`md-outlined-select` with `md-select-option`. Some of the surrounding code is changed, such as the default role in the user form, as part of the conversion.

## How it was tested

I went through each page that used these components and tested them—using Safari and Chrome on macOS. I did more careful testing where logic around the components changed (user form) and element types (tasks).

### New user form:
image

### New task form:
image

### New task form (open):
image

### Edit event form (the problematic one from #1363):
image

---

- [X] I have read [CONTRIBUTING.md](https://github.com/gramps-project/gramps-web/blob/main/CONTRIBUTING.md)
- [X] This was discussed in an issue first, or it is small enough not to need one
- [X] I wrote this myself, or used an AI assistant and have reviewed every line

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,6 @@
         "@material/mwc-linear-progress": "^0.27.0",
         "@material/mwc-list": "^0.27.0",
         "@material/mwc-menu": "^0.27.0",
-        "@material/mwc-select": "^0.27.0",
         "@material/mwc-snackbar": "^0.27.0",
         "@material/mwc-textfield": "^0.27.0",
         "@material/mwc-top-app-bar": "^0.27.0",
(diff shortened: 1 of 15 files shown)


DavidMStraub:
A heroic effort, thanks!

Issue #507, opened 2024-10-09
Feature request: Provide the ability in Grampsweb to change the genealogical symbols


Learnandgrow:
Currently, there is no known way to change genealogical symbols in Grampsweb.

This feature request is to provide the ability in Grampsweb to change the genealogical symbols.

Preferred if through the Grampsweb UI, but if quicker or as an interim step, allow the admin (or owner?) to update/specify the symbols to use in a configuration file.

dahar188:
Yes, please🙏

DavidMStraub:
If anyone wants to implement this, here is a rough guide.

- the choice of genealogical symbols should be stored using `this.appState.updateSettings` - similar to the language or home person
- a dropdown select needs to be added to `GrampsViewSettings.js` (user settings)
- note: the default should be the current behaviour
- throughout the code, the explicit symbols have to be replaced by the ones from the selected set

DavidMStraub:
Since per-tree config has been added in the meantime, it should also be evaluated whether this should instead be a per-tree config (by tree owners).

Probably, it makes sense to allow both: default < per-tree setting < per-user setting

yutongc4:
Hi @DavidMStraub, I’m a CodePath AI301 student interested in contributing to Gramps Web.

I’m interested in working on this issue. From your guidance, my initial understanding is that the selected genealogical symbol set should be stored through `this.appState.updateSettings`, similar to language or home person, and that a dropdown should be added to `GrampsViewSettings.js`.

Before starting, I had two quick questions:

1. Are the alternative symbol sets already defined somewhere in the codebase, or should I add an initial set of options?
2. For the first implementation, would you prefer starting with a per-user setting, or should it include the full priority order of default < per-tree setting < per-user setting?

I will keep the default behavior unchanged.

DavidMStraub:
1. No
2. You can start with per user

It's also worth reviewing how this is done in Gramps desktop - I don't remember if it's using predefined sets to choose from or allows to override each individual symbol, which would be way more flexible, but more complicated UI.

Nazivon:
Hello! I'm an Information Systems student, and I have a university project that involves making a significant contribution to an open-source project. I'd like to contribute to Gramps Web, as I've always been interested in genealogy.

Since yutongc4 doesn't seem to be working on this issue anymore, I'd be happy to take it on. Alternatively, if you think there are other issues that are also suitable for a beginner, I'd be glad to work on those afterwards. @DavidMStraub

Thank you!

[...]

PR #1382, opened 2026-09-06
Allow changing genealogical symbols

Nazivon:
## Linked issue

Issue #507: Providing the ability in Gramps Web to change the genealogical symbols

## What this changes

Genealogical symbols for birth, death, marriage, and divorce can now be changed from Settings > Appearance. The available options allow users to replace the pictorial symbols with their corresponding abbreviations ("b.", "d.", "m.", and "div.", respectively).

Additional symbol options may be added in the future, as well as the possibility of customizing each symbol individually.

## How it was tested

Tests were added to verify the behavior of the changes and were successfully run using:

- npm run test
- npx vitest symbols (for the newly created symbols test file)
- npx vitest settingsUser (for the settings test file)

The screenshots below show the previous birth and death symbols (an asterisk and a cross, respectively) being replaced with the corresponding abbreviations ("b." and "d.") through the new settings options.

image
image
image

---

- I have read [CONTRIBUTING.md](https://github.com/gramps-project/gramps-web/blob/main/CONTRIBUTING.md)
- This was discussed in an issue first: #507 .
- I wrote this myself, with the help of an AI assistant (Claude Sonnet 5) and have reviewed every line.


Nazivon:
There were some YAML-related issues involving the `package-lock.json` and `package.json` files, which caused the check to fail. I attempted to resolve them in the latest commits, and the check is now passing.

DavidMStraub:
Thanks for this!

Abbreviations like "b." need to be translated but I don't see an addition to `strings.js`. Please check if it actually works as they use `msgctxt` in Python which is currently ignored by the JS translation.

Nazivon:
> Thanks for this!
> 
> Abbreviations like "b." need to be translated but I don't see an addition to `strings.js`. Please check if it actually works as they use `msgctxt` in Python which is currently ignored by the JS translation.

You're welcome. Yes, I will review this and the things remarked by Copilot and will get back to it. Thank you!

Nazivon:
The changes recommended have been implemented, as well as I could manage. Should anything else be necessary, please inform!

DavidMStraub:
Sorry if my refactoring of the tree views caused additional work!

Nazivon:
> Sorry if my refactoring of the tree views caused additional work!

No, it's okay! It's all part of the process. I'll take some time to rework things and get back to the issue. Thank you!

Nazivon:
Hello, I’ve made a few changes since the last refactoring. When you have a chance, I’d really appreciate any feedback or suggestions to help make sure everything looks good. Thank you!

DavidMStraub:
Thanks!

Here's a review by Claude Opus, I think the points make sense:

---

Overall this is a clean, well-factored approach. `getSymbols(settings, translate)` in a single module is the right shape, and a grep for the hardcoded symbols comes back clean except for the two default literals noted below. Four things worth raising:

## 1. Visual regression on the person page

`src/components/GrampsjsPerson.js:127-130` swaps `asteriskIcon`/`crossIcon` (inline SVGs) for a text symbol but keeps the `<i>` wrapper. `src/SharedStyles.js:195-202` styles only `.event i svg`:

```css
.event i svg path { fill: var(--grampsjs-body-font-color-40); }
.event i svg { height: 16px; width: auto; }
```

Both rules now match nothing. The birth/death markers render as *italic* text at full body colour instead of the muted 40% at a normalised 16px. This needs an `.event i` rule (non-italic, muted colour) or a different wrapper element.

## 2. `renderPersonDates` gap in the tree chart

`src/components/GrampsjsTreeChart.js:189` still calls `renderPersonDates(person.profile)` with no options, so that panel keeps Unicode symbols regardless of the setting. The same file spreads `getSymbols(...)` into the chart update at line 143, so it is inconsistent within one component. `GrampsjsChildren` and `GrampsjsConnectedParents` were updated; this call site was missed.

## 3. Defaults duplicated and inconsistent

- `src/charts/personCard.js:46` defaults `birthSymbol = '*'` (ASCII U+002A), while everything else uses `'∗'` (U+2217).
- `src/components/personListUtils.js:37` hardcodes `'∗'`/`'†'` again.

Both should pull from `SYMBOL_SETS[SYMBOL_SET_DEFAULT]` in `src/symbols.js` so there is one source of truth.

## 4. Scope creep: the a11y/tabindex changes

`src/components/GrampsjsPillToggle.js:185-197` and `src/components/GrampsjsTable.js:304-309` change `'0'`/`'-1'` to numbers *and* add `eslint-disable-next-line lit-a11y/tabindex-no-positive`. Those two fixes are redundant with each other, and both are unrelated to genealogical symbols. If the lint failure is real it belongs in its own PR; if the number change alone fixes it, the disable comments should go.

## Minor

- `renderPerson` in `src/objectRender.js:22` has no callers anywhere in `src/` or `test/`, so the new `settings` parameter there is dead code. It is also the only reason `asteriskIcon`/`crossIcon` were dropped from that import, leaving them as unused exports in `icons.js`.
- `getSymbols` is called per row inside render loops (for example, per child in `GrampsjsChildren`), allocating a fresh object each time. Negligible in practice, but hoisting it out of the map would be cheaper.
- The `en.json` option labels embed the literal symbols in the message key (`"Unicode symbols (∗ † ⚭ ⚮)"`), which is awkward for Weblate. A neutral key with the sample rendered separately would age better.

The `b.`/`d.`/`m.`/`div.` additions to `src/strings.js` are handled correctly as backend Gramps strings rather than frontend `lang/en.json` keys.

Issue #1381, opened 2026-09-06
[Feature] Family tree timeline showing lifespans, overlaps, and historical context


sharst:
### What problem would this solve?

I've wanted something like this for a long time. Right now, understanding how two relatives' lives overlapped, how old someone was when their children were born, or how a family's history sits against wider historical events, means piecing dates together in your head across separate person and event pages. A shared visual timeline would make all of that legible at a glance - who was alive at the same time as whom, and how an individual life relates both to the rest of the family and to events outside it.

### Possible solution (optional)

Inspired by MyHeritage's Family Tree Timeline and by TimeNets, a visualization technique someone raised on the gramps-users mailing list back in 2010 (no further movement since): each ancestor drawn as a single lifespan bar on one shared calendar, generations stacked, bars coloured by lineage to match the existing fan chart's four-colour scheme, historical period bands (e.g. the World Wars) for context, and small on-bar markers for births, marriages, and deaths.

I put together an interactive prototype to make the idea concrete, styled to match Gramps Web's own colours and fonts: https://claude.ai/code/artifact/0eb5b4c2-32bf-4557-8473-d50831276d79

Generation depth is adjustable (2-4 generations), the chart can be zoomed and panned along the time axis, and clicking a bar shows that person's details. Names and dates in it are invented for demonstration.

Drafted with an AI assistant on my behalf.

- [x] I have searched for existing issues before creating this one
- [x] I wrote this myself, or used an AI assistant and have read the whole text and agree with it

DavidMStraub:
I assume you're suggesting an improvement of the existing timeline view?

sharst:
Actually I was planning this as a separate view - the existing timeline serves a different purpose of giving an overview of all events condensed on a single timeline. This here would allow to inspect individual family members and their relations in more detail. 

Am 6. September 2026 17:57:18 MESZ schrieb David Straub ***@***.***>:

DavidMStraub:
The question is where to put it - we could turn the timeline view into a view with multiple charts like the tree view. But a general challenge with these life span charts is that they need a filter to be feasible, like a family filter or a degree of separations filter.

Another option would be to treat it as a tree view, which would make sense if it shows the same people shown in a relationship chart...

sharst:
Following your thought, I would definitely put it next to the tree views. When thinking of it, it essentially is just an ancestor tree where each ancestor bubble happens to have a length proportional to the person's age. 

[...]

PR #1389, opened 2026-09-09
Prototype: Lifespan Timeline chart for the Family Tree view

sharst:
## Summary
This is an early, working prototype of the Lifespan Timeline feature requested in #1381 — built to get visual confirmation on whether this is the right direction before investing further.

Each ancestor is drawn as a single lifespan bar on a shared calendar, grouped into generation clusters (Parents / Grandparents / Great-grandparents / ...), colour-coded by sex to match the other tree charts, with markers for marriages and children's births, and gravestone/age markers at the end of each bar. It reuses the existing ancestor-tree data fetching, zoom/pan, click-to-navigate, and hover-preview conventions from the other tree charts.

## Known rough edge
The dashed parent→child connector lines can get visually busy, especially with more generations loaded — open to ideas on how to calm that down (e.g. only drawing them on hover/selection, or a different visual language entirely).

## Feedback wanted
This is meant as a discussion starting point, not a finished feature. Any input — on the generation-clustering approach, the overall visual language, what's missing, or whether this is worth continuing — very welcome.

## Test plan
- [ ] Open a person's Family Tree view → "Lifespans" tab
- [ ] Verify generation clusters, connector lines, markers, and zoom/pan
- [ ] `npm run typecheck` / `npm test` pass

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -419,5 +419,6 @@
   "New Blog Post": "New Blog Post",
   "Blog Post": "Blog Post",
   "Failed to fetch the Blog tag": "Failed to fetch the Blog tag",
-  "Only the most recent revisions are shown.": "Only the most recent revisions are shown."
+  "Only the most recent revisions are shown.": "Only the most recent revisions are shown.",
+  "Lifespans": "Lifespans"
 }
--- a/src/charts/LifespanChart.js
+++ b/src/charts/LifespanChart.js
@@ -1,0 +1,530 @@
+// Lifespan (pedigree timeline) chart: each ancestor as a single lifespan bar
+// on one shared calendar, clustered by generation (parents, grandparents,
+// great-grandparents, ...), colour-coded by sex to match the other tree
+// charts, with small on-bar markers for the birth of a child, marriages, and
+// death — using the same event-type colours as the Timeline view
+// (schemeSet1) and the app's own sex colours elsewhere.
+
+import {create} from 'd3-selection'
+import {scaleTime} from 'd3-scale'
+import {axisBottom} from 'd3-axis'
+import {zoom as d3zoom, zoomIdentity} from 'd3-zoom'
+import {schemeSet1} from 'd3-scale-chromatic'
+import {mdiGraveStone, mdiAccountOutline, mdiBabyCarriage, mdiInfinity} from '@mdi/js'
+import {getPersonByGrampsId, getTree} from './util.js'
+import {chartNameDisplayFormat} from '../util.js'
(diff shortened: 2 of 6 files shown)


DavidMStraub:
Awesome start!

I think there are still a few bugs regarding ages and also translations, but the bars are pretty IMO!
In the example below you also see that the vertical lines overlap when husband and wife are born at similar dates. I don't have a good UI idea how to solve this, or the issue you mention.

grafik

I think we shouldn't display any bars for "estimated" or "about" dates as these have ±50 years by default in Gramps. Not sure what to do for people where only birth *or* death is known, what do you think?

For the timeline (the axis), I would make it consistent with the timeline chart in style and font size. The fonts look overall quite small.

On mobile, the chart becomes unusable - we should think about some kind of graceful degradation.

grafik

sharst:
First of all, thanks for being so open and responsive on this feature request. 
So far it's really just meant as a visual prototype and to bring up questions exactly like yours, and not at all for looking at the code (yet). 

To answer your questions, I suggest the following terms:
* A **bar** is showing the lifespan of a person on the time axis
* A **line** is the connector between parents and children at the birth date
* The **timeline axis** is the x-axis with the years as tick marks. 

Now regarding your questions:
1. Lines overlap when husband and wife are born at similar dates: I would just keep it as is, as it's the most accurate representation and doesn't clutter visually. 
2. For estimated and about-dates I would propose to fade in the bar from [date - 25y] to [date + 25y]. 
2a. When both birth and death date are estimated, I would fade in from [birthdate - 25y] to [center between birth- and deathdate] and fade out from there until [deathdate + 25y]

3a. When the birthdate is not known but the deathdate is known, I would fade in from [deathdate -50y]
3b. When the deathdate is not known but the birthdate is known, I would fade out until [birthdate + 50y]
3c. When both the deathdate and birthdate are not known, I would not display any bar (if birthdates of children are known, the dashed line will connect from the child's birthdate to the empty space of the parent, which will still give a clear indication IMO.


4. Regarding the broken mobile view - I'd love to keep compatibility with smaller screens. One proposal could be that we distribute names and bars a bit differently; instead of
[Last Name] [Surname]   [Bar]
[Birthdate, Deathdate]
We could do:
[Last Name] [Surname] [Birthdate] [Deathdate]
[Bar]
This way the Bar could span the entire screen width, which should make the experience a bit better. 

Regarding fonts and sizes, agreed, I'll adapt them to the timeline view. 

What do you think? I'll wait for your input before I'll give claude another go ;)

PR #1433, opened 2026-09-15
Replace the tree view tabs with a segmented chart switcher

DavidMStraub:
Replaces the 85px tab bar of the tree view with a compact segmented chart switcher.

- Extends `GrampsjsPillToggle` without changing its existing uses:
  - optional icons per option
  - an opt-in `muted` style with a light primary tint for the selected option
  - an opt-in `icons-only-narrow` mode that shows only icons when the toggle is narrower than its labels need
  - `--grampsjs-pill-toggle-margin` and `--grampsjs-pill-toggle-padding`
- The chart switcher uses the muted style at 40px height. It switches charts with one click and with the arrow keys, and the chart's own keys ignore it.
- The charts get the height the tabs no longer need.
- The fit button uses the filled icon, matching the other toolbar icons.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


DavidMStraub:
Ready

PR #1436, opened 2026-09-16
Remove the unused mwc-menu and mwc-list packages

DavidMStraub:
The last `mwc-menu` was the tree chart's menu of relatives, which now uses `md-menu`, and no template renders `mwc-list` or `mwc-list-item` either. Their imports and style rules were left behind.

- Removes the `@material/mwc-menu` imports from five files and the `@material/mwc-list` imports from six.
- Removes the style rules for `mwc-list` in `SharedStyles.js` and `GrampsJs.js`, which matched nothing.
- Removes `mwc-list-item` from the elements where keyboard shortcuts are ignored.
- Removes both packages from `package.json` and the lockfile, which drops 504 lines of lockfile entries.

Lint, the unit tests and the production build pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


DavidMStraub:
#541

PR #1437, opened 2026-09-16
 Fix translation key typos so existing translations are used 

anthonyroussel:
## Linked issue

Closes #

## What this changes

Some translations weren't showing up translated in the Web interface because the translation key in the JS source code didn't exactly match the translation key registered in the translation files (like wrong case or typo). So the translation lookup helper always missed the translation and fell back to English, for every language:

* 'Add Y-DNA Data' -> 'Add Y-DNA data' (from Gramps Web)
* 'Move up' / 'Move down' -> 'Move Up' / 'Move Down' (from Gramps Desktop)
* 'Publication Info' -> 'Publication info' (from Gramps Desktop)
* 'Abbrevation' -> 'Abbreviation' (from Gramps Desktop)
* 'unknown' -> 'Unknown' (from Gramps Desktop)

## How it was tested

---

- [x] I have read [CONTRIBUTING.md](https://github.com/gramps-project/gramps-web/blob/main/CONTRIBUTING.md)
- [ ] This was discussed in an issue first, or it is small enough not to need one
- [x] I wrote this myself, or used an AI assistant and have reviewed every line


anthonyroussel:
Screenshots below show the strings now display translated:

### Abreviation + Publication info
image

### Move up
image

### Move down
image

### Unknown

image

### Add Y-DNA data

image

lang/en.json
```diff
@@ -110,6 +110,7 @@
   "login": "login",
   "Lost password?": "Lost password?",
   "Register new account": "Register new account",
+  "All": "All",
```
DavidMStraub: This is an existing backend string - this should go to `strings.js`

DavidMStraub:
Good catch! Clearly, this code was written in prehistoric times when fallible humans were using keyboards to assemble the source code letter by letter :rofl: 

Please just move "All" to strings.js.

lang/en.json
anthonyroussel: Indeed, "All" is a Gramps Desktop string.

I looked into moving it to strings.js, but it's not safe yet: I checked the actual Gramps Desktop .po catalogs, and for 7 of them (ar, id, ko, lv, nn, pt_BR, vi) "All" is untranslated, so it'd silently fall back to English.

And these 7 languages already had a working translation in lang/{lang}.json, so the move would've been a regression for them.

I've submitted the missing translations on the Gramps Desktop Weblate project, but that needs a maintainer review cycle on Weblate first.

I propose to revert back to keeping it in lang/en.json for now, and to follow up with a separate issue/PR to make the strings.js move once those translations land in the repository.

What do you think?

DavidMStraub: It's still the right thing to do.

I investigated and found the following history:

- 2022 "All" added to en.json, not present in Gramps 5.1 strings
- 2024 Gramps 5.2 released including a new "All" string
- 2025 "All" removed from the search view in #834, thus removed from en.json
- 2026 "All" added back in the user list, forgot to add it to strings.js

Yes, it's probably going to be a minor translation regression in some languages, but I don't think that justifies doing it incorrectly.

anthonyroussel: Thanks! I will revert to add "All" translation key to strings.js, then remove the "All" translations from the Gramps Web translation files.

anthonyroussel: Done

DavidMStraub:
Please don't remove the key from the language files except en.json - this would break Weblate sync AFAIK.

PR #1456, opened 2026-09-22
feat(settings): add browser push notification support

### LINKED ISSUE #1386: [Feature] Add Web Push client support
### What problem would this solve?

Gramps Web currently has no browser-side support for receiving Web Push notifications.

Calendar subscriptions or email notifications can cover some use cases, but they do not provide native browser notifications.

### Possible solution (optional)

Add generic Web Push client support:

* register a Service Worker;
* request notification permission;
* create subscriptions with PushManager;
* send/remove subscriptions through the API;
* display notifications;
* handle notification clicks;
* handle unsupported browsers and denied permissions.

### Before submitting

- [x] I wrote this myself, or used an AI assistant and have read the whole text and agree with it

elig-45:
## Summary

Add generic browser-side Web Push support to Gramps Web.

## Changes

- add Browser notifications under Settings > User for Gramps Web API 3.23+
- request notification permission only after an explicit user action
- create, synchronize, and remove the current browser subscription
- handle unsupported browsers, denied permissions, unavailable server configuration, and retryable failures
- display incoming notifications through the service worker
- focus or open a safe same-origin target when a notification is clicked
- add unit tests for subscription management and Web Push helpers

## Backend dependency

Requires [gramps-web-api#1001](https://github.com/gramps-project/gramps-web-api/pull/1001).

## Validation

- 13 targeted Web Push tests pass
- complete test suite passes: 669 tests
- typecheck, Prettier check, targeted ESLint, and production build pass
- the global ESLint command still reports two unrelated pre-existing accessibility errors

Closes #1386


lang/en.json
```diff
@@ -38,6 +38,22 @@
...
+  "Notification permission was not granted. You can try again.": "Notification permission was not granted. You can try again.",
+  "Browser notifications are enabled on this device.": "Browser notifications are enabled on this device.",
+  "Browser notifications are disabled on this device.": "Browser notifications are disabled on this device.",
+  "Browser notification settings could not be loaded.": "Browser notification settings could not be loaded.",
+  "The server subscription could not be removed.": "The server subscription could not be removed.",
+  "The browser subscription could not be removed.": "The browser subscription could not be removed.",
+  "Enable notifications": "Enable notifications",
+  "Enabling...": "Enabling...",
+  "Disable notifications": "Disable notifications",
+  "Disabling...": "Disabling...",
```
DavidMStraub, line 56: Do we really need 16 new strings for one feature??Perhaps some of it could just be icons?

src/views/GrampsjsViewSettingsUser.js
```diff
@@ -184,6 +185,21 @@ export class GrampsjsViewSettingsUser extends GrampsjsView {
         ${this.renderTreePreferences()}
       </grampsjs-collapsible-section>
 
+      ${this._supportsWebPush()
+        ? html`
+            <grampsjs-collapsible-section
```
DavidMStraub, line 190: The sections are meant to group related settings.

When adding a new one, it helps to think about what else might be added in the future, and be more generic. Otherwise we have stale translation keys everytime we add something new.

In this case, I suggest to use wording that would allow us to add email notifications, for instance, under the same heading/description.

elig-45:
Thanks, agreed. I’ll simplify this into a generic Notifications section, with Browser as the current channel so future options such as e-mail can live alongside it.

I’ll also reduce the browser-specific strings and use a more compact, accessible state control, keeping explicit text only where it is needed to explain an error or unavailable permission/configuration.

